PR #22, opened 2026-02-24
Improve DAG performance: cache Node.dependsOn, fix WorkQueue thread s…

stalep:
…afety

- Cache Node.dependsOn() ancestor IDs in a transient Set for O(1) lookups after first computation (was recursive DFS on every call, 5x upload speedup)
- Remove dead foundNode computation and add short-circuit in Work.dependsOn()
- Replace CopyOnWriteArrayList with ArrayList in KahnDagSort and WorkQueue
- Throw IllegalArgumentException on DAG cycles instead of silently reordering
- Replace PriorityQueue with ArrayDeque in isCircular()
- Guard WorkQueue reads with takeLock for thread safety
- Add unique constraints on node_edge and value_edge join tables
- Remove @Immutable from Work entity to allow retryCount mutations
- Fix Node.hashCode() to exclude sources (prevents StackOverflowError)
- Fix lazy init hacks in NodeService and ValueService
- Replace stream().anyMatch/filter/allMatch with plain for-loops in Work.dependsOn(), Work.equals(), Work.getAncestorWorks(), WorkQueue.getScore(), and WorkQueue.getRequiredPrecedingRunnables(). Eliminates Spliterator, Pipeline, and lambda capture allocations on every call. Upload time reduced from 56s to 48s (13% improvement).


stalep:
Here are the improvements I've observed using the `perf_test` benchmark.
```
  ┌───────────────────┬─────────────┬─────────────┐
  │      Version      │ Upload time │   vs main   │
  ├───────────────────┼─────────────┼─────────────┤
  │ main              │ 4m52s       │ baseline    │
  ├───────────────────┼─────────────┼─────────────┤
  │ + dependsOn cache │ 56s         │ 5.2x faster │
  ├───────────────────┼─────────────┼─────────────┤
  │ + stream removal  │ 48.6s       │ 6.0x faster │
  └───────────────────┴─────────────┴─────────────┘
```
The biggest win was to introduce caching.

With the removal of streams I got a 7.5% improvement which I thought was significant enough to include it.

Here's a list of the top hotspots according to the profiling I did:
``` ┌──────┬──────────────────────────────┬──────────────────────────┬───────────────────────────────┐
  │ Rank │     Main (4m52s)     │     With cache (56s)     │ With cache + no streams (48s) │
  ├──────┼──────────────────────────────┼──────────────────────────┼───────────────────────────────┤
  │ 1    │ Node.dependsOn 16,758        │ Method.invoke 368        │ ProxyInterceptor 356          │
  ├──────┼──────────────────────────────┼──────────────────────────┼───────────────────────────────┤
  │ 2    │ Node.getSources 4,611        │ ProxyInterceptor 332     │ WQ.getRequiredPreceding 350   │
  ├──────┼──────────────────────────────┼──────────────────────────┼───────────────────────────────┤
  │ 3    │ Node.hibernate_sources 4,032 │ Pipeline.copyInto 324    │ Node$Proxy.dependsOn 290      │
  ├──────┼──────────────────────────────┼──────────────────────────┼───────────────────────────────┤
  │ 4    │ Work.dependsOn 3,139         │ Pipeline.wrapAndCopy 279 │ Work.dependsOn 278            │
  ├──────┼──────────────────────────────┼──────────────────────────┼───────────────────────────────┤
  │ 5    │ Node.getId 3,114             │ MatchSink.accept 270     │ LazyInterceptor 252           │
  └──────┴──────────────────────────────┴──────────────────────────┴───────────────────────────────┘
```

h5m-core/src/main/java/io/hyperfoil/tools/h5m/entity/Node.java
```diff
@@ -162,45 +169,40 @@ public boolean equals(Object o){
 
     @Override
     public int hashCode(){
-        return Objects.hash(id, name, operation, List.copyOf(sources));
+        return Objects.hash(id, name, operation);
```
willr3: if `id==null` because it isn't persisted then I think we need to use the sources to differentiate two nodes with the same name and operation but different input nodes? Is the issue the List.copy or the use of each source in the hashCode? 
We ideally should not see very many nodes without an id so perhaps this will not be a concern.

willr3 (approved):
Everything looks good with just 2 comments / questions:
1. The comment about Node.hashCode no longer using sources
2. Do we need to use fetch join or can Hibernate / Panache be configured to do that with attribute annotations?

In either case I think it's safe to merge

stalep:
> Everything looks good with just 2 comments / questions:
> 
>     1. The comment about Node.hashCode no longer using sources

Yes, it's a valid concern. Before this change the `hashCode()` and `equals()` were not consistent, but they are now. In reality, I'm not sure it will be an issue as unpersisted nodes are persisted and then used in maps/sets. The window where an unpersisted node lives in a hash structure is narrow. The safest fix would be to add `sources.size()` to both `hashCode()` and `equals()` for unpersisted case. It would avoid the recursion risk we had with `List.copyOf`. 

>     2. Do we need to use fetch join or can Hibernate / Panache be configured to do that with attribute annotations?

The annotation options are:
  - `@Fetch(FetchMode.JOIN)` — makes it eager everywhere, causing unwanted joins in queries that don't need sources loaded
  - `@BatchSize(size=25)` — we already added this; it batch-loads lazily, which is a good default
  -` @EntityGraph` — more verbose, requires named graph definitions

  JOIN FETCH in specific queries gives surgical control — we only eagerly load where we know we need it, without affecting other code paths. The @BatchSize annotations we already added handle the general case.

PR #25, opened 2026-02-24
Switch to self-contained `jq4j`

andreaTP:
Switches `calculateJqValues()` from forking `/usr/bin/jq` via `ProcessBuilder` to using `jq4j`, which runs `jq` as a WASM module inside the JVM.                                                       
                                                            
  - No external `jq` binary required
  - No temp files
  - Removes "Install jq" step from CI workflows

  Existing `NodeServiceTest.calculateJqValues*` tests pass unchanged.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,9 +18,6 @@ jobs:
           distribution: 'temurin'
           cache: 'maven'
 
-      - name: Install jq
-        run: sudo apt-get update && sudo apt-get install -y jq
-
       - name: Build and test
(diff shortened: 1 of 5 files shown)


andreaTP:
Thanks for all the performance considerations!

I implemented a "reactor-mode" and an "instance pool" to re-use as much as possible the runtime instances.
A quick correction on "WASM interpreter overhead" we are already using `jq` compiled at build time with Chicory, there is no interpreter overhead, is the `jq` execution itself that is heavy.

Update coming soon!

PR #27, opened 2026-02-25
Structured fingerprints and jq-based fingerprint filter for change de…

stalep:
…tection

Replace SHA-256 hash fingerprints with structured JSON objects using sorted keys (TreeMap) so individual fingerprint components are inspectable and comparable. Add fingerprintFilter field to RelativeDifference that accepts a jq expression to gate which fingerprints participate in change detection.

### DIFF
--- a/h5m-core/src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
+++ b/h5m-core/src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
@@ -15,7 +15,6 @@ import io.hyperfoil.tools.h5m.entity.Value;
 import io.hyperfoil.tools.h5m.entity.node.*;
 import io.hyperfoil.tools.h5m.pasted.ProxyJacksonArray;
 import io.hyperfoil.tools.h5m.pasted.ProxyJacksonObject;
-import io.hyperfoil.tools.yaup.hash.HashFactory;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 import jakarta.persistence.EntityManager;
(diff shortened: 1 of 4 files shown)


h5m-core/src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
```diff
@@ -833,17 +836,52 @@ public static ObjectNode convertMapping(org.graalvm.polyglot.Value value) {
...
+        try {
+            File inputFile = Files.createTempFile("h5m.fp.filter.", ".json").toFile();
+            inputFile.deleteOnExit();
+            ObjectMapper mapper = new ObjectMapper();
+            Files.write(inputFile.toPath(), mapper.writeValueAsBytes(fingerprint));
+
+            ProcessBuilder processBuilder = new ProcessBuilder(
+                    "/usr/bin/jq",
+                    "--exit-status",
+                    filter,
```
willr3: We should not pass the user provided filter as a parameter to the process as I worry it is a security issue where a malicious / malformed filter could cause unintended code execution. The `calculateJqValues` uses `--from-file` to store the filter in a file.

Looking at the production Horreum database, the fingerprint_filter is a javascript method that needs to return a `truthy` value (if the filter exists ) for the fingerprint to be used in the calculation. Do we want to keep that implementation or port the existing JS to jq?

willr3 (commented):
I do not think we should merge in the current state and we should decide if we are changing the implementation of fingerprint filtering.

h5m-core/src/test/java/io/hyperfoil/tools/h5m/svc/NodeServiceTest.java
```diff
@@ -731,4 +730,281 @@ public void calculateValues_NxN_complicated() throws SystemException, NotSupport
...
+        relDifference.setMinPrevious(1);
+        relDifference.setFingerprintFilter("(fp) => fp.platform === \"x86\"");
+        relDifference.setNodes(fingerprintNode,rootNode,rangeNode,domainNode);
+
+        List<Value> found = nodeService.calculateRelativeDifferenceValues(relDifference,rootX86_01,0);
+        assertNotNull(found);
+        // verify that any results come only from x86 fingerprints
+        for (Value v : found) {
+            assertNotNull(v.data, "change detection result should have data");
+        }
```
willr3: This appears to be verifying that the Value.data is not null but it is not verifying that the Values are from the x86 related Values. Do you see it differently?

PR #29, opened 2026-02-26
Replace external jq binary with jackson-jq for in-process filter eval…

stalep:
…uation

Replaces ProcessBuilder/temp-file jq invocation with jackson-jq (net.thisptr:jackson-jq:1.6.0), a pure Java jq implementation built on Jackson. This eliminates the system dependency on an installed jq binary, removes process forking and file I/O overhead, and simplifies CI workflows.

- Add jackson-jq dependency to parent and h5m-core pom.xml
- Rewrite NodeService.calculateJqValues() to use JsonQuery.compile/apply in-process
- Add JsonQuery compilation cache (ConcurrentHashMap) to avoid redundant compilations
- Harden isNullInput() regex with word boundaries to avoid false matches
- Harden inputs/[inputs] string replacement with regex word boundaries
- Remove dead ValueService.writeToFile() and its test (only used by old ProcessBuilder approach)
- Remove unused imports from ValueService, ValueServiceTest, NodeService
- Remove 'Install jq' step from CI and pull-request workflows
- Update README and design docs to reflect in-process jq evaluation

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,9 +18,6 @@ jobs:
           distribution: 'temurin'
           cache: 'maven'
 
-      - name: Install jq
-        run: sudo apt-get update && sudo apt-get install -y jq
-
       - name: Build and test
         run: mvn -B clean verify
 
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -18,9 +18,6 @@ jobs:
           distribution: 'temurin'
           cache: 'maven'
 
-      - name: Install jq
-        run: sudo apt-get update && sudo apt-get install -y jq
-
       - name: Build and test
         run: mvn -B clean verify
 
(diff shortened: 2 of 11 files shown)


willr3:
I am worried about the correctness of using something that emulates the `jq` interface. I was testing the branch and I get an error condition for a test I added to my copy of your branch. 

https://github.com/willr3/h5m/pull/new/jackson_jq

I ran the same test using `jq`

```
▷ cat one.json 
{ "value": 10, "foo": { "value" : 1} }

▷ cat two.json 
{ "value": 20, "foo": { "value" : 2} }

▷ jq --null-input --slurp "inputs | map(.value) | add" one.json two.json 
30

▷ jq --version
jq-1.7.1
```
I worry about us encountering situations where user's have data mutations working with `jq` that they cannot port to h5m due to differences in the implementation. 
Horreum suffers from a similar problem now where the jsonpath implementation is tied to postgresql and therefore can only be tested with an instance of postgresql.

Issue #14, opened 2026-02-04
Change notification


willr3:
h5m needs a way to provide feedback when the change detection nodes (currently just RelativeDifference) detect a change. The implementation should work for both the current cli based interface but also support asynchronous notifications (e.g. email and slack).

willr3:
I am trying to understand the value of phase 2 and all I can come up with is the ability to confirm and add a description to the Value from the change detection node. What does confirming and describing provide to users?
```
Change {
    detectionNode  → Node      // which detection node (rd, ft, ed)
    changeValue    → Value     // the Value with change details
    rootValue      → Value     // the upload that triggered detection
    folder         → Folder    // for efficient querying
    fingerprint    : String    // for matching across recalculations
    confirmed      : Boolean   // null=unreviewed, true=confirmed, false=dismissed
    description    : String    // user triage notes
    timestamp      : LocalDateTime
}
```
The `detectionNode`, `changeValue`, `rootValue`, `folder`, and `fingerprint`  call all be found from the Value DAG. What do we gain by duplicating the information into another table at this point in the tool design?

PR #37, opened 2026-03-10
Fire CDI event on change detection (#14)

stalep:
## Summary

- Fire a `ChangeDetectedEvent` CDI event when a detection node (FixedThreshold, RelativeDifference) detects a change during work execution
- Add `NodeType.isDetection()` to classify detection node types
- Add a transient `notify` flag to `Work` so recalculations don't trigger change detection events — uploads keep the default `notify=true`, `FolderService.recalculate()` sets `notify=false`
- Cascade work inherits the parent's `notify` flag

## Test plan

- `ChangeDetectionTest` — 4 test cases:
  - Event fires when a threshold is violated
  - Event does not fire when value is within threshold
  - Event does not fire when `notify=false` (recalculation)
  - Event does not fire for non-detection nodes

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/api/NodeType.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/api/NodeType.java
@@ -22,4 +22,5 @@ public enum NodeType {
     public String display() {
         return display;
     }
+
 }
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,10 +11,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Set up Java 21
+      - name: Set up Java 22
         uses: actions/setup-java@v4
         with:
-          java-version: '21'
+          java-version: '22'
           distribution: 'temurin'
           cache: 'maven'
 
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -11,10 +11,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Set up Java 21
+      - name: Set up Java 22
         uses: actions/setup-java@v4
         with:
-          java-version: '21'
+          java-version: '22'
           distribution: 'temurin'
           cache: 'maven'
 
--- a/pom.xml
+++ b/pom.xml
@@ -57,7 +57,7 @@
         <!-- dependencies -->
         <lib.duckdb>1.4.4.0</lib.duckdb>
         <lib.jdbi>3.41.3</lib.jdbi>
-        <lib.jjq>0.1.1</lib.jjq>
+        <lib.jjq>0.1.2</lib.jjq>
         <lib.jmh>1.37</lib.jmh>
         <lib.jsonata>2.6.2</lib.jsonata>
         <lib.mapstruct>1.6.3</lib.mapstruct>
@@ -290,7 +290,6 @@
                     <release>${maven.compiler.release}</release>
                     <compilerArgs>
                         <arg>-parameters</arg>
-                        <arg>--enable-preview</arg>
                     </compilerArgs>
                     <annotationProcessorPaths>
                         <path>
@@ -362,7 +361,7 @@
                 <artifactId>maven-surefire-plugin</artifactId>
                 <version>${build.maven-surefire-plugin}</version>
                 <configuration>
-                    <argLine>@{argLine} --enable-preview</argLine>
+                    <argLine>@{argLine}</argLine>
                     <excludes>**/*IT.java</excludes>
                     <excludes>%regex[.*IT.class]</excludes>
                     <systemPropertyVariables>
--- a/src/main/java/io/hyperfoil/tools/h5m/entity/NodeEntity.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/entity/NodeEntity.java
@@ -132,6 +132,10 @@ public abstract class NodeEntity extends PanacheEntity implements Comparable<Nod
 
     public abstract NodeType type();
 
+    public boolean isDetection() {
+        return type() == NodeType.FIXED_THRESHOLD || type() == NodeType.RELATIVE_DIFFERENCE;
+    }
+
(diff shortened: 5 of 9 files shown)


willr3 (commented):
I think the PR includes changes that go beyond the description and we should split them to discuss their implications. I do not think the "idle" support is a good plan for h5m but would be interested in discussing them in a separate PR.

stalep:
Simplified the pr, can you have a look @willr3 ?

src/main/java/io/hyperfoil/tools/h5m/api/svc/FolderServiceInterface.java
```diff
@@ -49,14 +51,16 @@ public interface FolderServiceInterface {
...
+
+    List<ValueEntity> getDetectionValues(String name, List<ValueEntity> rootValues);
 
     /**
      * Recalculates the contents or state of a folder by its name.
      *
      * @param name The name of the folder to recalculate.
      */
-    void recalculate(String name);
+    List<Long> recalculate(String name);
```
willr3: What is being returned by the recalculate method?

stalep:
@willr3 can you have another look?

src/main/java/io/hyperfoil/tools/h5m/queue/WorkQueue.java
```diff
@@ -31,15 +31,65 @@ public class WorkQueue implements BlockingQueue<Runnable> {
...
+        RootValueTracker(Condition done) {
+            this.count = 0;
+            this.done = done;
+        }
+    }
+
+    private Long getRootValueId(Work work) {
+        return (work.sourceValues != null && !work.sourceValues.isEmpty())
+                ? work.sourceValues.get(0).getId() : null;
+    }
```
willr3: What is the expected behavior when there are more than 1 source values? Why is only the first value pertinent?

src/main/java/io/hyperfoil/tools/h5m/api/svc/WorkServiceInterface.java
```diff
@@ -17,4 +17,6 @@ public interface WorkServiceInterface {
      */
     boolean terminate(long timeout, TimeUnit timeUnit) throws InterruptedException;
 
+    boolean awaitWork(long rootValueId, long timeout, TimeUnit unit) throws InterruptedException;
```
willr3: I am strongly opposed to adding a blocking api that will turn a threadpool in h5m into a blocking task executor. I believe we can find a less wasteful way to accomplish the goal once we discuss and agree upon the desired functionality for the users.

src/main/java/io/hyperfoil/tools/h5m/queue/WorkQueue.java
```diff
@@ -31,15 +31,65 @@ public class WorkQueue implements BlockingQueue<Runnable> {
...
 
+    private static class RootValueTracker {
+        int count;
+        final Condition done;
+
+        RootValueTracker(Condition done) {
+            this.count = 0;
+            this.done = done;
+        }
+    }
```
willr3: The `RootValueTracker` appears to be added to facilitate notification that all work is complete for a specific Value instance. We already have that functionality with the KahnDag sorting of the workQueue. I think we should identify what user facing functionality we are adding and evaluate if an existing component already provides the necessary backend features.

willr3 (commented):
I do not think we should rely on a blocking design for notification that the scope of work for a specific Value has completed. I instead favor using the sorting workQueue to add a Runnable that is sorted to occur after all other Runnables that pertain to the same Value. Perhaps a `Work` that is not persisted and lacks an `activeNode`? Using a blocking api will lead to a thread being held idle while work is performed and we thus limit the capacity of the process to the size of that threadpool.

stalep:
Thanks for the thorough review @willr3, here's what I'm thinking:

**Agreed**: We should not add blocking/idle-wait APIs to the WorkQueue. The `RootValueTracker` adds complexity to a critical concurrent data structure and turns the thread pool into a blocking executor —that was added because of the cli.

**Proposed alternative**: Replace the `RootValueTracker`/`awaitRootValue` mechanism with a **CDI event** fired when the last Work in a DAG subtree completes. Something like:

```java
@ApplicationScoped
public class WorkCompletionEvent {
    // fired when all downstream Work for a root value completes
}
```

- WorkService fires the event after deleting the last Work item for a given root value
- The CLI observes the event synchronously to get upload results before exiting
- Later, the web UI can observe the same event to push notifications via SSE/WebSocket

This avoids any blocking in the WorkQueue, leverages existing CDI infrastructure, and gives us one notification mechanism that serves both CLI and web UI consumers.

Re: the other feedback:
- `noWait` → agree on renaming to `wait` or removing it
- `byNameEager` → will consolidate with existing `byName`
- `recalculate` return value → will be clearer once the completion event replaces the root value ID plumbing
- Entities in service API → agree, should return DTOs not JPA entities

stalep:
@willr3 can you have another look. The pr is simplified and should only fire a cdi event when a change has been detected.

src/main/java/io/hyperfoil/tools/h5m/api/NodeType.java
```diff
@@ -22,4 +22,8 @@ public enum NodeType {
     public String display() {
         return display;
     }
+
+    public boolean isDetection() {
```
barreiro: I think this method should belong to `Node`, not `NodeType`

barreiro (changes requested):

src/main/java/io/hyperfoil/tools/h5m/api/NodeType.java
stalep: I changed it to be on Node in pr#59 :)

willr3:
We discussed on the team call that the `notify` flag on `Work` should not be necessary. The `FolderService.recalculate` should be changed to only queue the top level Nodes (node's that list the `root` as a source) so that recalculation only occurs for nodes when new source values are available. After that I think it this parallel CDI event based notification is fine for notifications that do not need a guarantee of delivery like we have for Work execution.

Issue #5, opened 2026-01-18
Closure table delete entity cascade


willr3:
Deleting a node with references in a closure table needs to delete all closure table entries for that node but not entries that other nodes also provide. If A -> X and B-> X then removing A from a node's sources should not remove X if the node also has B as a source

willr3:
fixed in willr3/closure branch

PR #44, opened 2026-04-07
fix edge table delete cascade to preserve shared children (#5)

stalep:
When deleting a node/value with references in edge tables (node_edge, value_edge), shared children are now preserved if they have other parents. Fixes the entity/query mismatch in getDependentValues, adds parent count checks before cascading deletes, and cleans up inverse edge rows.


willr3:
This PR claims to fix closure table delete cascade in the main branch but I don't see where the main branch is using a closure table and this PR does not re-introduce a closure table.

stalep:
yeah, the fix itself is about preserving shared children in these edge tables during delete cascades (checking parent count before removing), but the PR title/description should have said "edge table" instead of "closure table". Updating now.

Issue #4, opened 2026-01-17
Test sqlpath when no value is found


willr3:
Need a unit test for when sqlpath and sqlpathall do not match data from the input value. H5m does not use null value.data

willr3:
sqlpathall is silently missing data compared to other nodes which leaves the user unable to identify which iteration is missing and therefore they cannot create any valid datasets / comparisons. This is impacting the production Horreum as well as the PoC.

stalep:
@willr3 I added a pr documenting the current behavior. Should we rather change h5m to use null objects?

Arjun-Parmani:
Tests calculateSqlJsonpathValues_null() and calculateSqlAllJsonpathValues_null() covers this issue.

PR #45, opened 2026-04-07
add tests for sqlpath and sqlpathall no-match behavior (#4)

stalep:
Test that sqlpath, sqlpathall, and partial-match scenarios correctly return no values when the jsonpath expression does not match input data.


stalep:
@willr3 should we change the current behavior or keep it as it is you think?

willr3:
@stalep I believe it is correct for the Node to fail to produce a Value in the scenario where it cannot extract the target data from the input Value(s).

PR #46, opened 2026-04-07
Add REST API with JAX-RS/OpenAPI annotations and endpoint tests

stalep:
Make h5m ready for a React frontend by adding REST endpoints to existing service interfaces.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -162,6 +162,18 @@
             <artifactId>quarkus-picocli</artifactId>
         </dependency>
         <dependency>
+            <groupId>io.quarkus</groupId>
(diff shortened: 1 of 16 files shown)


barreiro (changes requested):
I think the REST endpoints should be defined in their own classes. otherwise the CLI will have to pull in all the web dependencies. 

My plan was to have `*Resource` classes in a new package where the endpoints are defined, leaving the `*ServiceInterface` as normal java interfaces.

stalep:
I agree with your suggestion @barreiro! I updated the pr, now I created the endpoints in separate classes in the rest folder. Can you have another look?

PR #48, opened 2026-04-07
Fix work queue race conditions, add benchmark infrastructure

stalep:
## Problem

Rapid back-to-back uploads produce `StaleObjectStateException` and silently drop computed values. The work queue reports itself as idle while worker threads are still processing, causing test cleanup (and potentially production monitoring) to act on stale state.

## Root Cause

`ExecutorConfiguration` produced `WorkQueueExecutor` with `@Dependent` CDI scope — **every `@Inject` site got its own executor with its own WorkQueue**. `WorkService` submitted work to one queue while tests (and any other consumer) checked `isIdle()` on a completely different, always-empty queue.

This was invisible at runtime because the system appeared to work — work *was* being processed. But any code that relied on `isIdle()` (test cleanup, monitoring, future SSE completion events) operated on the wrong queue instance.

```
WorkService  →  @Inject WorkQueueExecutor  →  instance A (has work)
Test class   →  @Inject WorkQueueExecutor  →  instance B (always empty, isIdle() = true)
```

### 1. Single executor ownership (root cause)

Deleted `ExecutorConfiguration`. `WorkService` now creates the `WorkQueueExecutor` internally during `@Observes StartupEvent`, making the thread pool a private implementation detail. No CDI scope ambiguity is possible — there's exactly one queue because there's exactly one `WorkService` (`@ApplicationScoped`).

Tests access idle state through `workService.isIdle()` instead of injecting the executor directly.

**Files:** `WorkService.java` (owns executor), `ExecutorConfiguration.java` (deleted), `WorkServiceInterface.java` (`isIdle()` added), 3 test files (inject `WorkService`)

### 2. Deferred queue insertion until transaction commits

`WorkService.create()` persists Work entities via `em.merge()` + `em.flush()`, but the DB row isn't visible to other transactions until the outer transaction commits. Previously, work was added to the in-memory queue immediately — worker threads could pick it up before the row was visible, causing `StaleObjectStateException`.

Fix: register a JTA `Synchronization` callback that adds work to the queue in `afterCompletion(STATUS_COMMITTED)`. A `deferredCount` on `WorkQueue` ensures `isIdle()` returns false while deferred items are waiting for commit.

**Files:** `WorkService.java` (afterCompletion in create/execute), `WorkQueue.java` (deferredCount)

### 3. equals/hashCode contract fixes

- **`Work.equals()`** only compared `sourceNodes` when `activeNode == null`, but `hashCode()` always included them. Two Work items could have equal hashCodes but unequal equals — breaking HashSet/HashMap behavior in `pendingWork`/`activeWork`.
- **`Node` API record** had default record `equals`/`hashCode` that recursed infinitely through `group → sources → group → ...`. Added ID-based short-circuit.

**Files:** `Work.java`, `Node.java`

## Benchmark Infrastructure

Added `EdgeTableBenchmarkTest` and `UploadPipelineBenchmarkTest` with `BenchmarkTimer` and `DbGraphBuilder` utilities. These are excluded from `mvn test` by default (in `src/test/java/.../benchmark/`). Also includes:

- `@BatchSize(100)` on edge collections (was 25, mismatched global setting)
- `parent_id` indexes on `node_edge` and `value_edge` (2.4x speedup for recursive CTEs)
- `KahnDagSort` short-circuit when no internal dependencies exist (3.5x faster for flat topologies)
- Early return in `NodeEntity.computeAncestorIds()` for leaf nodes


willr3:
I have serious doubts about the code in this PR as it claims to benchmark closure tables but I cannot find where a closure table is used in the main branch and I don't see where the PR is introducing a closure table

stalep:
Updating the PR title, description, and the benchmark class/doc references.

willr3:
The batch size is being set to 100 in NodeEntity and ValueEntity which matches the new default value being set in application.properites. Do we want to move forward with configuration overrides in NodeEntity and ValueEntity?
```
quarkus.hibernate-orm.fetch.batch-size=100
```
```
@BatchSize(100)
```

PR #49, opened 2026-04-10
Created test cases for threshold change detection

Arjun-Parmani:


src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
```diff
@@ -210,10 +210,12 @@ public void calculate_relativedifference_node(QuarkusMainLauncher launcher) thro
...
         assertTrue(last.getOutput().contains("Count: 13"),"expect 13 values from test");
     }
-    @Test //not yet working because relativedifference doesn't know about the "datsaet" node
+
+    @Test
+    //not yet working because relativedifference doesn't know about the "datsaet" node
     //need to tell relativedifference which node is the dataset. either detect it with CTE (is that possible)
-    //or make it an attribute on the FolderEntity / NodeGroupEntity
-    @Disabled("There should be only changes detected for x = 2 and x = 12 but there are two other detected for x = 3 and x = 13")
+    //or make it an attribute on the FolderEntity / NodeGroup
```
willr3: I don't think this needs to be changed to NodeGroup

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
```diff
@@ -210,10 +210,12 @@ public void calculate_relativedifference_node(QuarkusMainLauncher launcher) thro
...
+    @Disabled("There should be only changes detected for x = 2 and x = 12 but there are two other detected for x = 3 and x=13")
```
willr3: Please don't change to an inconsistent spacing

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
```diff
@@ -226,16 +228,20 @@ public void calculate_relativedifference_dataset_node(QuarkusMainLauncher launch
                   "each": [
                     {"x": 3, "y": 1.1, "fp1": "alpha", "fp2": "alpha"},
                     {"x": 13, "y": 20.2, "fp1": "alpha", "fp2": "bravo"}
+             
                   ]
                 }
                 """
         );
+
+
```
willr3: extra spacing does not seem necessary to separate files

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
```diff
@@ -226,16 +228,20 @@ public void calculate_relativedifference_dataset_node(QuarkusMainLauncher launch
                   "each": [
                     {"x": 3, "y": 1.1, "fp1": "alpha", "fp2": "alpha"},
                     {"x": 13, "y": 20.2, "fp1": "alpha", "fp2": "bravo"}
+
```
willr3, line 230: does this extra space in the string literal serve a purpose?

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
```diff
@@ -244,8 +250,8 @@ public void calculate_relativedifference_dataset_node(QuarkusMainLauncher launch
                 """
                 {
                   "each": [
-                    {"x": 1, "y": 3.1, "fp1": "alpha", "fp2": "alpha"},
-                    {"x": 11, "y": 40.2, "fp1": "alpha", "fp2": "bravo"}
+                   {"x": 1, "y": 3.1, "fp1": "alpha", "fp2": "alpha"},
+                   {"x": 11, "y": 40.2, "fp1": "alpha", "fp2": "bravo"}
```
willr3, line 250: Did you mean to change the indentation of the string or was this unintentional?

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
```diff
@@ -226,16 +228,20 @@ public void calculate_relativedifference_dataset_node(QuarkusMainLauncher launch
...
         );
+
+
         Path filePath02 = Files.writeString(Files.createTempFile(folder,"h5m",".json").toAbsolutePath(),
                 """
                 {
                   "each": [
                     {"x": 2, "y": 2.1, "fp1": "alpha", "fp2": "alpha"},
                     {"x": 12, "y": 30.2, "fp1": "alpha", "fp2": "bravo"}
+
```
willr3: blank spaces in a string literal should not be necessary

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
```diff
@@ -258,19 +264,22 @@ public void calculate_relativedifference_dataset_node(QuarkusMainLauncher launch
                 new String[]{"add","jq","to",testName,"fp1","{split}:.fp1"},
                 new String[]{"add","jq","to",testName,"fp2","{split}:.fp2"},
                 new String[]{"list",testName,"nodes"},
-                new String[]{"add","relativedifference","relativediff","to",testName,"range","rangeNode","domain","domainNode","by","split","fingerprint","fp1,fp2","window","1","minPrevious","1"},
+
+                new String[]{"add","relativedifference","relativediff","to",testName,"range","rangeNode","domain","domainNode","by","split","fingerprint","fp1,fp2","window","1","minPrevious","1","threshold","0.1"},
```
willr3: why did you need to change the threshold for this test? I thought this PR was focused on the threshold testing

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
```diff
@@ -258,19 +264,22 @@ public void calculate_relativedifference_dataset_node(QuarkusMainLauncher launch
...
                 new String[]{"list",testName,"nodes"},
                 new String[]{"upload",filePath01.toString(),"to",testName},
                 new String[]{"upload",filePath02.toString(),"to",testName},
                 new String[]{"upload",filePath03.toString(),"to",testName},
                 new String[]{"list","value","from",testName}
+
+
```
willr3: why was this space added?

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
```diff
@@ -744,6 +753,195 @@ public void list_values_as_table_group_by(QuarkusMainLauncher launcher) throws I
...
+        assertTrue(last.getOutput().contains("\"y\":70.0,\"fp1\":\"alpha\""),
+                "Appropriate fingerprint matching failed: Alpha should be tied to 70.0");
+
+        //checking for appropriate fingerprint matching
+        assertTrue(last.getOutput().contains("\"y\":175.0,\"fp1\":\"beta\""),
+                "Appropriate fingerprint matching failed: Beta should be tied to 175.0");
+        assertTrue(last.getOutput().contains("\"y\":150.0,\"fp1\":\"beta\""),
+                "Appropriate fingerprint matching failed: Beta should be tied to 150.0");
+        assertTrue(last.getOutput().contains("\"y\":100.0,\"fp1\":\"beta\""),
+                "Appropriate fingerprint matching failed: Beta should be tied to 100.0");
```
willr3: These asserts can fail if the order of `y` and `fp1` changes in the toString for the json becuase they rely on matching a full substring rather than ensuring the line with a specific `y` value has the correct `fp1` value. It would also be good if you include the incorrect line in the error message for the assert. That allows us to see what happened during the failing test rather than having to change the test to output the value and re-run it.

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
```diff
@@ -744,6 +753,195 @@ public void list_values_as_table_group_by(QuarkusMainLauncher launcher) throws I
...
+        assertTrue(last.getOutput().contains("\"y\":70.0,\"fp1\":\"alpha\""),
+                "Appropriate fingerprint matching failed: Alpha should be tied to 70.0");
+
+
+        assertTrue(last.getOutput().contains("\"y\":175.0,\"fp1\":\"beta\""),
+                "Appropriate fingerprint matching failed: Beta should be tied to 175.0");
+        assertTrue(last.getOutput().contains("\"y\":150.0,\"fp1\":\"beta\""),
+                "Appropriate fingerprint matching failed: Beta should be tied to 150.0");
+        assertTrue(last.getOutput().contains("\"y\":100.0,\"fp1\":\"beta\""),
+                "Appropriate fingerprint matching failed: Beta should be tied to 100.0");
```
willr3: please see the above comment about the asserts, the same thing applies here.

willr3 (changes requested):
Please also squash these two commits into a single commit as the scope of this PR is 2 new tests and ideally that can be done in a single commit.

stalep:
## Fix for FixedThreshold fingerprint scoping

The tests in this PR revealed a real bug in `calculateFixedThresholdValues` — I've pushed a fix to branch `fix-threshold-fingerprint-scoping` that makes all the new tests pass.

### What was broken

The old code had three problems:

1. **Only used the first fingerprint** — the `break` after finding the first matching fingerprint meant all other fingerprints were ignored, and all violations were tagged with that one fingerprint
2. **Range values not scoped by fingerprint** — used `getDescendantValues(root, rangeNode)` which returned ALL range values under the root regardless of which fingerprint they belong to
3. **Incorrect ancestor linkage** — used `getAncestor(root, groupBy)` instead of `getAncestor(fingerprintValue, groupBy)`

### The fix

- **Iterates all fingerprints** (removed the `break`) matching the pattern used by `calculateRelativeDifferenceValues` and Horreum's per-fingerprint processing
- **Scopes range values per fingerprint** using `findMatchingFingerprint()` with the nearest common ancestor node of the fingerprint and range nodes as the boundary. A new `findScopingNode()` method walks the node dependency graph to find this boundary automatically — so it works correctly both with and without an explicit `by` parameter
- **Correct ancestor linkage** — violations are linked via `getAncestor(fingerprintValue, groupBy)`

### Test results

All 112 tests pass (including both new tests from this PR):
- `calculate_fixedthreshold_with_multiple_parent_values` ✅  
- `calculate_fixedthreshold_with_multiple_parent_values_with_by_split` ✅

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
Arjun-Parmani: removed the extra spacing

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
Arjun-Parmani: removed extra spacing

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
Arjun-Parmani: The indentation was unintentional. I was trying to change the order of file upload for testing and while reverting the change I believe I had an extra space here.

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
Arjun-Parmani: Same reason for this as well. The indentation was unintentional. I was trying to change the order of file upload for testing and while reverting the change I believe I had an extra space here.

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
Arjun-Parmani: Removed extra space. While testing locally I must have added an extra blank line unintentionally.

stalep:
closing, replaced by pr#59

Issue #54, opened 2026-04-16
Improve Work retry logic and error handling


stalep:
## Context

Split from #51 to keep the race condition fix PR focused.

## Description

Improve how `WorkService.execute()` handles failures: increase retry limit, persist retry state, and use structured logging.

### Changes needed

- Increase `RETRY_LIMIT` from 0 to 3
- Persist `retryCount` on the managed entity so retry state survives restarts
- Delete permanently failed Work (exceeding retry limit) instead of leaving it in DB
- Defer retry re-queue via `afterCompletion` to avoid re-queuing before the retryCount update is visible

### Why

With `RETRY_LIMIT=0`, any transient failure permanently drops work. Retries allow recovery from transient DB errors, connection timeouts, etc. Structured logging makes failures searchable and includes work/node IDs for debugging.

PR #56, opened 2026-04-17
Bump retry limit to 3 and delete permanently failed work (#54)

stalep:
Previously RETRY_LIMIT was 0 (no retries) and work that exceeded the limit was left in the database indefinitely. Now failed work retries up to 3 times with structured logging, and is deleted after exhausting retries to prevent unbounded accumulation.


willr3:
I think it is helpful to delay this until we have more of h5m implemented as retry and losing failed work makes it difficult to debug correctness.

stalep:
Closing for now

Issue #52, opened 2026-04-13
Security & Roles Design


stalep:
## Summary

h5m needs authentication and authorization to support multi-user deployments. This issue tracks the design and implementation of a simplified security model compared to Horreum.

### Key Design Decisions

- **Two deployment modes**: local (no auth, all users are admin) and service (OIDC + API keys)
- **OIDC provides authentication only** — all roles and team memberships managed in h5m's database
- **Two role types**: global admin and team member (binary membership)
- **Folder is the ownership boundary** — every folder belongs to a team, all other entities inherit team context
- **Read access is fully public** — only write/upload/modify operations require authorization
- **No PostgreSQL RLS** — application-level enforcement via a single `AuthorizationService`
- **API keys** for CLI authentication, configurable expiration (default 1 year)
- **Team self-management** — any team member can add/remove members on their team

### New Entities

- `Team` — name (unique), members (`@ManyToMany` with User)
- `User` — username (from OIDC), role (admin/user), teams
- `ApiKey` — SHA-256 hash, expiration, linked to User

### Implementation Phases

1. Entity model + deployment mode configuration
2. OIDC + API key authentication
3. Authorization enforcement (`AuthorizationService`)
4. REST API + HTTP security annotations

### Full Design Document

[docs/design/plan-security-roles.md](https://github.com/Hyperfoil/h5m/blob/main/docs/design/plan-security-roles.md)

PR #58, opened 2026-04-19
Add security and roles infrastructure (issue #52)

stalep:
Introduces User, Team, and ApiKey entities with AuthorizationService as a single authorization checkpoint. Security is config-driven via h5m.security.enabled (default false) preserving backward compatibility. Adds CLI admin commands for user/team management and tests for all new services including security-enabled profile tests.

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/entity/ApiKey.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/entity/ApiKey.java
@@ -1,0 +1,47 @@
+package io.hyperfoil.tools.h5m.entity;
+
+import io.quarkus.hibernate.orm.panache.PanacheEntity;
+import jakarta.persistence.Column;
+import jakarta.persistence.Entity;
+import jakarta.persistence.FetchType;
+import jakarta.persistence.ManyToOne;
+
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+
+@Entity(name = "api_key")
+public class ApiKey extends PanacheEntity {
+
+    @Column(name = "key_hash")
+    public String keyHash; // SHA-256 hex
+
+    @ManyToOne(fetch = FetchType.LAZY)
+    public User user;
+
+    public String description;
+
+    public Instant createdAt;
+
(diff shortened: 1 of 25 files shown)


src/main/java/io/hyperfoil/tools/h5m/entity/ApiKey.java
```diff
@@ -0,0 +1,33 @@
...
+    public String keyHash; // SHA-256 hex
+
+    @ManyToOne(fetch = FetchType.LAZY)
+    public User user;
+
+    public String description;
+
+    public Instant createdAt;
+
+    public Instant expiresAt;
```
barreiro: I think it's better to have a last used date (and a number of idle days) and compute the expiration

see https://github.com/Hyperfoil/Horreum/blob/master/horreum-backend/src/main/java/io/hyperfoil/tools/horreum/entity/user/UserApiKey.java

src/main/java/io/hyperfoil/tools/h5m/entity/User.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+@Entity(name = "h5m_user")
+public class User extends PanacheEntity {
+
+    @Column(unique = true)
+    public String username;
+
+    public String role; // "admin" or "user"
```
barreiro: This assumes the roles won't overlap and a users have a single role in H5M. Given the experience we have with Horreum I think it's fine.

I rather see an enum here.

barreiro (commented):
Still needs some more work to use the features (implement `SecurityIdentityAugmentor` if we want to use `@RolesAllowed` annotation, implement `HttpAuthenticationMechanism`, etc. ) but for a start is good.

stalep:
> Still needs some more work to use the features (implement `SecurityIdentityAugmentor` if we want to use `@RolesAllowed` annotation, implement `HttpAuthenticationMechanism`, etc. ) but for a start is good.

Yeah, I did not want it to be a too large of a pr so for now it just adds the foundation.

PR #59, opened 2026-04-19
Added DetectionNode and fixed fingerprint scoping for all detection types

stalep:
Introduces a DetectionNode interface (getFingerprintNode, getGroupByNode, getRangeNode, getFingerprintFilter) implemented by FixedThreshold and RelativeDifference. The shared findScopingNode(DetectionNode) method computes the nearest common ancestor of fingerprint and range nodes via BFS, replacing the user-specified groupBy in findMatchingFingerprint calls. This fixes incorrect cross-fingerprint matching when a split node fans uploads into multiple items.

- FixedThreshold: scopes range values per fingerprint instead of fetching all
- RelativeDifference: uses scopingNode for domain/range matching
- NodeType.isDetection() added for generic detection node identification
- Static ObjectMapper constant replaces per-call instantiation
- Test helper extracted to reduce FixedThreshold test duplication
- Enables previously disabled relativedifference dataset test

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
@@ -507,62 +507,57 @@ public class NodeService implements NodeServiceInterface {
             List<ValueEntity> fingerprintValues = valueService.getDescendantValues(root, ft.getFingerprintNode());
             String fpFilter = ft.getFingerprintFilter();
 
-            // Check fingerprint filter — if all fingerprints are filtered out, skip this root
-            JsonNode fingerprintData = null;
             for (int fIdx = 0; fIdx < fingerprintValues.size(); fIdx++) {
                 ValueEntity fingerprintValue = fingerprintValues.get(fIdx);
                 if (fpFilter != null && !evaluateFingerprintFilter(fpFilter, fingerprintValue.data)) {
                     continue;
                 }
-                fingerprintData = fingerprintValue.data;
-                break;
-            }
-            if (fingerprintData == null) {
-                return rtrn;
-            }
 
-            // Get range values scoped to this root only
-            List<ValueEntity> rangeValues = valueService.getDescendantValues(root, ft.getRangeNode());
-            for (int rIdx = 0; rIdx < rangeValues.size(); rIdx++) {
-                ValueEntity rangeValue = rangeValues.get(rIdx);
-                Double numericValue;
-                if (rangeValue.data instanceof NumericNode numericNode) {
-                    numericValue = numericNode.asDouble();
-                } else if (rangeValue.data.toString().matches("[0-9]+\\.?[0-9]*")) {
-                    numericValue = Double.parseDouble(rangeValue.data.toString());
-                } else {
-                    continue;
-                }
-
-                FixedThreshold.ViolationType violationType = null;
-                double bound = Double.NaN;
-
-                if (ft.isMinEnabled()) {
-                    if (ft.isMinInclusive() ? numericValue < ft.getMin() : numericValue <= ft.getMin()) {
-                        violationType = FixedThreshold.ViolationType.BELOW;
-                        bound = ft.getMin();
+                // Get range values scoped to this fingerprint and current root
+                List<ValueEntity> rangeValues = valueService.findMatchingFingerprint(
+                    ft.getRangeNode(), groupBy, fingerprintValue, null, null, root, -1, -1, true
+                );
+                for (int rIdx = 0; rIdx < rangeValues.size(); rIdx++) {
+                    ValueEntity rangeValue = rangeValues.get(rIdx);
+                    Double numericValue;
+                    if (rangeValue.data instanceof NumericNode numericNode) {
+                        numericValue = numericNode.asDouble();
+                    } else if (rangeValue.data.toString().matches("[0-9]+\\.?[0-9]*")) {
+                        numericValue = Double.parseDouble(rangeValue.data.toString());
+                    } else {
+                        continue;
                     }
-                }
-                if (violationType == null && ft.isMaxEnabled()) {
-                    if (ft.isMaxInclusive() ? numericValue > ft.getMax() : numericValue >= ft.getMax()) {
-                        violationType = FixedThreshold.ViolationType.ABOVE;
-                        bound = ft.getMax();
-                    }
-                }
 
-                if (violationType != null) {
-                    ObjectNode data = OBJECT_MAPPER.createObjectNode();
-                    data.set("value", new DoubleNode(numericValue));
-                    data.set("bound", new DoubleNode(bound));
-                    data.put("direction", violationType.label());
-                    data.set("fingerprint", fingerprintData);
-                    ValueEntity changeValue = new ValueEntity(root.folder, ft, data);
-                    changeValue.idx = startingOrdinal;
-                    List<ValueEntity> foundParents = valueService.getAncestor(root, groupBy);
-                    if (foundParents.size() == 1) {
-                        changeValue.sources = foundParents;
+                    FixedThreshold.ViolationType violationType = null;
+                    double bound = Double.NaN;
+
+                    if (ft.isMinEnabled()) {
+                        if (ft.isMinInclusive() ? numericValue < ft.getMin() : numericValue <= ft.getMin()) {
+                            violationType = FixedThreshold.ViolationType.BELOW;
+                            bound = ft.getMin();
+                        }
                     }
-                    rtrn.add(changeValue);
+                    if (violationType == null && ft.isMaxEnabled()) {
+                        if (ft.isMaxInclusive() ? numericValue > ft.getMax() : numericValue >= ft.getMax()) {
+                            violationType = FixedThreshold.ViolationType.ABOVE;
+                            bound = ft.getMax();
+                        }
+                    }
+
+                    if (violationType != null) {
+                        ObjectNode data = OBJECT_MAPPER.createObjectNode();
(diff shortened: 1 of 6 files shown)


stalep:
This PR supersedes #49. It includes all the FixedThreshold scoping fixes and tests from that branch, generalized via a `DetectionNode` interface so the same `findScopingNode` logic applies to RelativeDifference (and future detection algorithms like Hunter). PR #49 can be closed once this merges.

src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
```diff
@@ -499,70 +498,91 @@ public List<ValueEntity> calculateRelativeDifferenceValues(RelativeDifference re
         return rtrn;
     }
 
+    /**
+     * Finds the nearest common ancestor node of the fingerprint and range nodes.
+     * This determines the correct scoping boundary for matching range values to fingerprints,
+     * even when the user-specified groupBy is broader (e.g., root).
+     */
+    private NodeEntity findScopingNode(DetectionNode detection) {
```
willr3: Could you please help me understand when this function would be appropriate? the `groupBy` node is the user defined node that represents the desired scoping boundary for the change detection. This function appears to be making assumptions about the `Node` / `Value` graph and those assumptions are not part of the model's constraints so they can be wrong.
We discussed this method during the team meeting last week and agreed that is represented flawed logic. Is there something new that makes this functionality appropriate where we previously thought it was not appropriate?

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
```diff
@@ -210,10 +209,7 @@ public void calculate_relativedifference_node(QuarkusMainLauncher launcher) thro
         LaunchResult last = results.getLast();
         assertTrue(last.getOutput().contains("Count: 13"),"expect 13 values from test");
     }
-    @Test //not yet working because relativedifference doesn't know about the "datsaet" node
-    //need to tell relativedifference which node is the dataset. either detect it with CTE (is that possible)
-    //or make it an attribute on the FolderEntity / NodeGroupEntity
-    @Disabled("There should be only changes detected for x = 2 and x = 12 but there are two other detected for x = 3 and x = 13")
+    @Test
```
willr3: Another example of removing comments and descriptive text. Please ensure this is not part of unrelated PRs. The test was disabled because there is a functional flaw in RelativeDifference change detection when the uploads are not in domainNode order. This PR does not appear to address that flaw.

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
```diff
@@ -226,16 +222,20 @@ public void calculate_relativedifference_dataset_node(QuarkusMainLauncher launch
                   "each": [
                     {"x": 3, "y": 1.1, "fp1": "alpha", "fp2": "alpha"},
                     {"x": 13, "y": 20.2, "fp1": "alpha", "fp2": "bravo"}
+
```
willr3: Why are we adding non-functional blank lines in string literals?

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
```diff
@@ -226,16 +222,20 @@ public void calculate_relativedifference_dataset_node(QuarkusMainLauncher launch
...
         );
+
+
         Path filePath02 = Files.writeString(Files.createTempFile(folder,"h5m",".json").toAbsolutePath(),
                 """
                 {
                   "each": [
                     {"x": 2, "y": 2.1, "fp1": "alpha", "fp2": "alpha"},
                     {"x": 12, "y": 30.2, "fp1": "alpha", "fp2": "bravo"}
+
```
willr3: another blank line in a string literal. This should not have a functional impact and so I think it is not appropriate with this PR.

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
```diff
@@ -258,19 +258,22 @@ public void calculate_relativedifference_dataset_node(QuarkusMainLauncher launch
                 new String[]{"add","jq","to",testName,"fp1","{split}:.fp1"},
                 new String[]{"add","jq","to",testName,"fp2","{split}:.fp2"},
                 new String[]{"list",testName,"nodes"},
-                new String[]{"add","relativedifference","relativediff","to",testName,"range","rangeNode","domain","domainNode","by","split","fingerprint","fp1,fp2","window","1","minPrevious","1"},
+                new String[]{"add","relativedifference","relativediff","to",testName,"range","rangeNode","domain","domainNode","by","split","fingerprint","fp1,fp2","window","1","minPrevious","1","threshold","0.1"},
```
willr3: why is a PR for threshold detection making a change to the relative difference testing? This should not be touched for this PR.

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
```diff
@@ -258,19 +258,22 @@ public void calculate_relativedifference_dataset_node(QuarkusMainLauncher launch
...
+
         );
         results.forEach(result->{
             assertEquals(0,result.exitCode(),result.getOutput());
         });
 
         LaunchResult last = results.getLast();
-        assertTrue(last.getOutput().contains("Count: 38"),"expect 38 values from test");
+        assertTrue(last.getOutput().contains("Count: 40"),
+                "expect 40 values from test\n" + last.getOutput());
```
willr3: Why was this change introduced in this PR? I do not understand why the value `38` is considered incorrect nor why why this scoped to be in this PR.

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
```diff
@@ -744,6 +747,96 @@ public void list_values_as_table_group_by(QuarkusMainLauncher launcher) throws I
...
+    }
 
+    @Test
+    public void calculate_fixedthreshold_with_multiple_parent_values_with_by_split(QuarkusMainLauncher launcher) throws IOException {
+        runFixedThresholdMultipleParentValues(launcher,
+                "calculate_fixedthreshold_with_multiple_parent_values_with_by_split",
+                new String[]{"range", "rangeNode", "by", "itemSplit", "fingerprint", "categoryFp", "min", "10", "max", "100"});
+    }
+
+    private void runFixedThresholdMultipleParentValues(QuarkusMainLauncher launcher, String testName, String[] ftArgs) throws IOException {
```
willr3: `calculate_fixedthreshold_with_multiple_parent_values` and `calculate_fixedthreshold_with_multiple_parent_values_with_by_split` call this method with different parameters but they do not have their own `assertX` validation so the different in their parameters do not have a meaningful impact on the test condition. Either the `private` method should become a standalone test or there should be something different in what is validated by the two test methods.

willr3 (changes requested):
I am concerned about the use of the `scopingNode` concept that appears to contradict the user provided `groupBy` Node. I think it relies on assumptions about the shape of the graph that are not part of the model. Why do you think we should introduce a scopingNode?

src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
stalep: Hi Will, appreciate the thorough review. To be transparent, with the changes I though it was logical to see if it would fix the relativedifference issue.

**`findScopingNode`:** You're right that this auto-infers scoping from the graph rather than respecting the user's `groupBy`. The intent was to handle the dataset/split case where `groupBy=root` causes cross-fingerprint contamination. What approach do you think would be more appropriate — requiring the user to set `groupBy` correctly for split scenarios, or adding a separate explicit parameter?

**RelativeDifference changes:** You're right — those changes should not be in this PR. I'll revert those changes and keep this PR focused on the FixedThreshold scoping fix and DetectionNode interface only.

stalep:
One thing I noticed: when `by` is omitted, the CLI silently defaults `groupBy` to root. For flat data this works fine, but for split scenarios (e.g., `.items[]`) how wound that work?
 Should we add validation that warns when the detection node's fingerprint/range sources have a common split ancestor when `by` isn't set, or is it reasonable to expect users to know they need by for splits?

willr3:
> For flat data this works fine, but for split scenarios (e.g., .items[]) how wound that work?

The `by` argument is necessary for users who have a splitting node that they wish to behave the same way `datasets` behave in the current Horreum model.

> is it reasonable to expect users to know they need by for splits?

I think it would be wrong to make any assumptions about the node graph and the user's intent. I do, however, think it would be valuable to detect when there are multiple fingerprint values, domain values, or range values for the target `byValue`. This would be an indication that the change detection configuration is not functioning properly. A `splitting` parent Node could be the cause but so could an error in Node operation or fingerprinting.

stalep:
> > For flat data this works fine, but for split scenarios (e.g., .items[]) how wound that work?
> 
> The `by` argument is necessary for users who have a splitting node that they wish to behave the same way `datasets` behave in the current Horreum model.
> 
> > is it reasonable to expect users to know they need by for splits?
> 
> I think it would be wrong to make any assumptions about the node graph and the user's intent. I do, however, think it would be valuable to detect when there are multiple fingerprint values, domain values, or range values for the target `byValue`. This would be an indication that the change detection configuration is not functioning properly. A `splitting` parent Node could be the cause but so could an error in Node operation or fingerprinting.

Thanks, that makes sense. The pr now rely entirely on the user-specified `groupBy` via `by`, no graph assumptions. It would be good to add logic to detect multiple fingerprint/domain/range values pr `byValue` group as a misconfiguration.

PR #63, opened 2026-04-23
fix infinite recursion in Node.hashCode

willr3:


stalep (commented):
The fix correctly identifies the infinite recursion through `Node` → `group` → `NodeGroup.sources` → `Node`, but I think there are a few remaining issues:

**1. `equals` still has the same recursion problem**

`hashCode` breaks the cycle by using `group.id()`, but `equals` still uses `Objects.equals(n.group, group)` which calls `NodeGroup`'s auto-generated `equals` — that includes `List<Node> sources`, which recurses back into `Node.equals`. Same stack overflow, different method.

**2. `sources` can also cause recursion**

Both `equals` and `hashCode` include `sources` via deep comparison/hashing (`Objects.hash(..., sources)` calls `List.hashCode()` → each element's `hashCode()`). If two nodes reference each other as sources (which the DAG shouldn't allow but the API records don't enforce), this would also overflow.

**3. Suggestion: match NodeEntity's approach**

`NodeEntity` already solves this — when `id != null`, `equals` compares by ID only and `hashCode` uses `Objects.hash(id, name, operation)`. For the `sources` hash it goes one level deep using source IDs rather than recursing. Since `Node` is the API projection of `NodeEntity`, it could follow the same pattern:

```java
@Override
public int hashCode() {
    if (id != null) {
        return Objects.hash(id, name, type);
    }
    return Objects.hash(name, type, operation, group != null ? group.id() : null);
}

@Override
public boolean equals(Object o) {
    if (o instanceof Node n) {
        if (id != null && n.id != null) {
            return id.equals(n.id);
        }
        return Objects.equals(name, n.name)
            && Objects.equals(type, n.type)
            && Objects.equals(operation, n.operation)
            && Objects.equals(group != null ? group.id() : null, n.group != null ? n.group.id() : null);
    }
    return false;
}
```

This avoids recursion in both methods and stays consistent with how the entity layer handles it.

Issue #57, opened 2026-04-19
OIDC and API key authentication (Phase 2 security)


stalep:
## Summary

Follow-up to #52 which added the entity model, AuthorizationService, and CLI admin commands. This issue covers wiring up actual authentication so the REST API enforces the security model.

## Scope

1. **OIDC integration** — add `quarkus-oidc` dependency and configure for Keycloak (or compatible provider). Provision h5m `User` entity on first login.
2. **Auto-admin bootstrap** — if the user table is empty when a new user logs in via OIDC, automatically create them as admin (replaces manual `h5m admin create-user` bootstrap).
3. **API key authentication** — implement `HttpAuthenticationMechanism` that validates `H5M_` prefixed keys against SHA-256 hashes in the `api_key` table. Wire to `SecurityIdentity`.
4. **REST security annotations** — protect write endpoints (upload, delete, recalculate) using `AuthorizationService`. Read endpoints remain public.

## Out of scope

- Per-team roles (viewer/tester/uploader/manager) — h5m uses binary membership
- Row-level security policies
- API key management REST endpoints (can be added later)

## References

- Design doc: `docs/design/plan-security-roles.md`
- Horreum reference: `ApiKeyAuthenticationMechanism`, `RolesAugmentor`, `ApiKeyIdentityProvider`
- Entity model already in place: `User`, `Team`, `ApiKey`, `FolderEntity.team`

PR #64, opened 2026-04-25
Add OIDC and API key authentication (issue #57)

stalep:
Wire up HTTP authentication for h5m's REST API. API keys use Authorization: Bearer H5M_xxx with SHA-256 hashing. OIDC is opt-in via environment variables. First user to authenticate is auto-promoted to admin. Read endpoints are public, write endpoints require auth, value purge requires admin role. Local mode provides a synthetic admin identity so existing behavior is unchanged.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -160,6 +160,10 @@
         </dependency>
         <dependency>
             <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-oidc</artifactId>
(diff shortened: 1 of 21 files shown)


barreiro (approved):
looks good.

probably should also add some `@RolesAllowed` annotations on the service implementations to enforce the authorization, but can be done at a later stage.

PR #69, opened 2026-04-29
Fix createParameters to wrap source values in ObjectNode for unmatche…

### LINKED ISSUE #67: createParameters passes raw value instead of named-property object for single-source functions
## Problem

`JsNode.createParameters()` has a shortcut at line 105: when `params.size() == 1 && sourceValues.size() == 1`, it passes the raw source value directly instead of wrapping it in a named-property ObjectNode.

This breaks functions that access properties by name, like merge nodes:
```javascript
obj => Object.values(obj).find(v => v != null)
```

When called with a single source value `"10000712.a848fd6d"`, `Object.values(string)` iterates the string's characters and returns `["1","0","0","0",...]`. The function then returns `"1"` instead of the full string.

## Where it happens

`JsNode.createParameters()` lines 104-106:
```java
if(params.size()==1 && sourceValues.size()==1){
    rtrn.add(sourceValues.get(sourceValues.keySet().iterator().next()).data);
}
```

## Expected behavior

When a function parameter name doesn't match any source value key, and there's exactly one source, the value should still be wrapped in an ObjectNode with the source name as key — matching what Horreum does for multi-extractor labels.

## Impact

Affects the legacy import's no-transform path where labels from multiple schema versions are merged with `obj => Object.values(obj).find(v => v != null)`. When only one schema version produces a value, the merge function receives a raw value instead of `{labelName: value}`, causing truncation.

Discovered during boot-time-nightly-rhivos (test 294) import — RHIVOS Build, RHIVOS Mode, OS Release, etc. all truncated to first character.

stalep:
…d single-param functions (#67)

When a JS function has a single parameter name that doesn't match any source value key (e.g., obj => Object.values(obj).find(v => v != null)), createParameters now wraps all source values into a named-property ObjectNode instead of returning an empty list.

Previously, the function would receive no arguments, causing Object.values(undefined) to throw or — when the raw value was passed through a different code path — Object.values("string") to iterate characters, truncating values to their first character.

Fixes #67

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/entity/node/JsNode.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/entity/node/JsNode.java
@@ -99,7 +99,11 @@ public class JsNode extends NodeEntity {
                 }
             }else{
                 if(!sourceValues.containsKey(param)){
-                    System.err.println("unable to find parameter value for " + param);
+                    if(params.size()==1 && !sourceValues.isEmpty()){
+                        ObjectNode obj = mapper.createObjectNode();
+                        sourceValues.forEach((k,v) -> obj.set(k, v.data));
+                        rtrn.add(obj);
+                    }
                 }else{
                     if(currentNode != null){
                         currentNode.set(param,sourceValues.get(param).data);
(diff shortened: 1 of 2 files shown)


stalep:
no need, fixed in pr#74

PR #71, opened 2026-04-29
Auto-split array values for element-by-element pairing in multi-sourc…

### LINKED ISSUE #66: Support per-iteration value pairing for multi-source labels
## Problem

When multiple `sqlall` extractors produce arrays from the same upload, downstream JS/JQ nodes receive the full arrays rather than individual elements paired by index. This prevents labels that combine per-iteration fields from different extractors.

**Example from rhivos-perf-comprehensive import:**

The Autobench Single Core label has two extractors:
- `results` → `$.results.*.SingleCore` (array of scores per iteration)
- `workload` → `$.workload` (string per iteration, e.g. "autobench")

The label function:
```javascript
value => {
    if (value["workload"] != "autobench") { return null }
    if (!value["results"]) { return 0 }
    return parseFloat(((value["results"].reduce((a,b) => a+b, 0)) / value["results"].length).toFixed(3))
}
```

In Horreum, the transformer splits the run into per-iteration datasets. Each dataset has one `workload` string and one `results` object. The label function processes each dataset independently.

In h5m, `sqlall` extractors return the full array as a single value. The label receives `workload = ["stressng", "fio", "coremark_pro", "autobench"]` instead of `workload = "autobench"`. The guard check always fails.

## Expected behavior

When a JS/JQ node has multiple `sqlall` sources that produce arrays of the same length, h5m should pair elements by index — `results[0]` with `workload[0]`, `results[1]` with `workload[1]`, etc. — and run the node function once per pair.

## Possible approaches

1. **Split node**: Add a split stage between `sqlall` extractors and label nodes that expands arrays into individual values. Downstream nodes then process one element at a time. This is analogous to Horreum's transformer splitting arrays into datasets.

2. **Indexed multi-iteration mode on JS/JQ nodes**: A node property that tells the execution engine to iterate source arrays by index instead of passing them whole. Similar to `MultiIterationType.Length` but applied to `sqlall` array contents.

3. **Implicit splitting in `calculateSourceValuePermutations`**: When source values are arrays from `sqlall` nodes, automatically expand them into per-element permutations.

## Affected labels (rhivos-perf-comprehensive)

- Autobench Single Core, Multi Core, Scaling, Results — guard on `$.workload`, process `$.results`
- Any future label that combines per-iteration fields from different `sqlall` extractors

## Context

This was discovered during the legacy import refactoring that eliminated the transformer/dataset layer (commits on `legacy-tests-fixes` branch). 32 of 36 labels work correctly with the direct-wiring approach. These 4 are the remaining gap.

stalep:
…e nodes (#66)

When calculateSourceValuePermutations sees multiple source nodes that each produced a single value containing a JSON array, it now expands the arrays into individual elements and pairs them by index.

This handles the case where sqlall extractors produce one ValueEntity with array data (e.g., [result0, result1, result2]) and a downstream JS node needs to process elements individually — pairing results[0] with workload[0], results[1] with workload[1], etc.

Previously the JS node received the full arrays as arguments, causing functions like value["workload"] != "autobench" to compare an array against a string, always failing.

Fixes #66

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
@@ -216,6 +216,29 @@ public class NodeService implements NodeServiceInterface {
             nodeValues.put(source.name, found);
         }
 
+        // Auto-split: when all sources have exactly 1 value but some contain JSON arrays,
+        // expand them into individual values for element-by-element pairing.
+        // This handles sqlall extractors whose single value is an array of N elements.
+        if (node.sources.size() > 1 && nodeValues.values().stream().allMatch(v -> v.size() == 1)) {
+            boolean hasArrayData = nodeValues.values().stream()
+                    .anyMatch(v -> v.get(0).data != null && v.get(0).data.isArray() && v.get(0).data.size() > 1);
+            if (hasArrayData) {
+                for (Map.Entry<String, List<ValueEntity>> entry : new ArrayList<>(nodeValues.entrySet())) {
+                    ValueEntity val = entry.getValue().get(0);
+                    if (val.data != null && val.data.isArray() && val.data.size() > 1) {
+                        List<ValueEntity> expanded = new ArrayList<>();
+                        for (int i = 0; i < val.data.size(); i++) {
+                            ValueEntity element = new ValueEntity(val.folder, val.node, val.data.get(i));
+                            element.idx = i;
+                            element.sources = val.sources;
+                            expanded.add(element);
+                        }
+                        nodeValues.put(entry.getKey(), expanded);
+                    }
+                }
+            }
+        }
+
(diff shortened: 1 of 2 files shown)


stalep:
Moved to draft based on Will's feedback.

Will is right — the auto-split is too aggressive. It changes behavior for ALL multi-source nodes with single array values, not just the import scenario. If a user intentionally stores an array as a single value (e.g., a list of measurements) and passes it to a JS function that handles array pairing internally, the auto-split would silently break their data by splitting it into individual elements.

The assumption "if all sources have single array values, they must need element-by-element pairing" is not safe — users should control whether arrays are split, not h5m making assumptions about data shape.

If this capability is needed in the future, it should be **opt-in** — either a node property (`multiType` or similar) or an explicit split node in the DAG — not automatic behavior in `calculateSourceValuePermutations`.

stalep:
Closing — the auto-split was solving a problem specific to the dataset-free import approach. With the transformer→dataset pipeline, array splitting happens naturally at the dataset level, matching Horreum's behavior. No core change needed.

PR #73, opened 2026-04-29
Add graph view filtering and label wrapping

stalep:
Improve the 'list nodes as Graph' command for large DAGs:
- Collapse coalesce nodes by default (hide _primary/_alt nodes)
- Add --type filter for comma-separated node types
- Add --depth limit for max levels from root
- Add --show-type flag to prefix labels with [type]
- Add --max-label-width to wrap long labels at word boundaries (default 30)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -53,7 +53,7 @@
         <h5m.benchmark.warmup>5</h5m.benchmark.warmup>
 
         <h5m.cli.native>true</h5m.cli.native>
-        <lib.aesh>3.5</lib.aesh>
+        <lib.aesh>3.6.1</lib.aesh>
 
         <!-- dependencies -->
         <lib.duckdb>1.4.4.0</lib.duckdb>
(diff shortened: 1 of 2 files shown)


willr3:
the `--type` filtering looks like it is implemented as graph truncation not filtering. The graph will stop walking at the point it encounters a `type` and will not include any nodes that are only linked to the graph by nodes of that `type`.  What node types do you find need to be removed from the graph and would it be better for the user to encode that information in some other way?

stalep:
> the `--type` filtering looks like it is implemented as graph truncation not filtering. The graph will stop walking at the point it encounters a `type` and will not include any nodes that are only linked to the graph by nodes of that `type`. What node types do you find need to be removed from the graph and would it be better for the user to encode that information in some other way?

Yes, it was truncating, not filtering.  now the --type filter now skips non-matching nodes but continues walking through them — children reconnect to the nearest visible ancestor instead of being truncated. Skipped nodes aren't added to the nodes map so they don't appear in the graph, but their sources are still walked so deeper matching nodes remain reachable.

PR #75, opened 2026-04-30
add identity check and improve parameter parsing for JsNode

willr3:
These changes were made to JsNode for the load-legacy-[test,run] work that is occurring in a separate branch but the changes should be shared while other work occurs on the load-legacy branch.


stalep:
Which (again) could be build on the pr for issue#68...

PR #76, opened 2026-04-30
Add verifyimport command to compare h5m data against Horreum

stalep:
CLI command that connects to both a Horreum legacy database and the h5m database, comparing:
- Node structure (label counts, change detection counts)
- Label values per dataset ordinal for each run
- Missing/extra/mismatched labels

Usage:
  h5m verifyimport testId=347 runId=42426 \ username=horreum-user password=horreum \ url=jdbc:postgresql://0.0.0.0:6000/horreum


stalep:
included in pr#102

PR #77, opened 2026-05-01
CalculateRelativeDifference only calculates change based on new upload

Arjun-Parmani:
The new implementation of NodeService.calculateRelativeDifferenceValues calculates the domain values based on the new upload (root id) for matching fingerprint values.

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
@@ -377,12 +377,12 @@ public class NodeService implements NodeServiceInterface {
             NodeEntity groupBy = NodeEntity.findById(relDiff.getGroupByNode().getId());
             List<ValueEntity> fingerprintValues = valueService.getDescendantValues(root,relDiff.getFingerprintNode());
             String fpFilter = relDiff.getFingerprintFilter();
-            for(int fIdx=0; fIdx<fingerprintValues.size(); fIdx++){
+            for(int fIdx=0; fIdx<fingerprintValues.size(); fIdx++) {
                 ValueEntity fingerprintValue = fingerprintValues.get(fIdx);
                 if (fpFilter != null && !evaluateFingerprintFilter(fpFilter, fingerprintValue.data)) {
                     continue;
                 }
-                if( relDiff.getDomainNode()!=null ){
+                if (relDiff.getDomainNode() != null) {
 
                     //when would this have more than 1 value?
                     /*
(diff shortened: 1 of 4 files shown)


stalep (commented):
Good work on the tests — they clearly demonstrate the expected behavior for both the recalculation and minPrevious scenarios.

A few things to consider:

**Domain ordering in the range query**

Setting `domainNode` and `domainValue` to `null` in the second `findMatchingFingerprint` call (line 417-418) removes the domain-based ordering for range values. The sliding window comparison (`window + minPrevious` values split into "previous" and "recent") depends on values being ordered by domain — without it, you might get values in an arbitrary order, and the window wouldn't represent the correct time-series segment.

Take a look at what `findMatchingFingerprint` does with `domainNode` and `domainValue` in `ValueService` — the domain node controls the ORDER BY, and the domain value controls the "preceding this point" cutoff. You want to keep that ordering but scope which domain points you iterate over.

**Hint**: The commented-out code block above your change (lines 388-414) was an earlier attempt at exactly this approach — it gets domain values scoped to the root, then for each one fetches range values with domain ordering preserved. That's the right structure. The reason it was commented out was the TODO about not handling values after previous detections, which is explicitly out of scope for this issue.

**Test assertions**

The value count assertions (`Count: 11`, `Count: 17`) verify the total number of values in the folder, which indirectly confirms change detection happened. A more direct assertion would check the actual change detection values — for example, verify the detection value contains the expected domain point or ratio. This would catch cases where the right number of values exist but the wrong changes were detected.

Arjun-Parmani:
On **Domain ordering in the range query**

Initially the comparison was done based on domain-value (v.data≤domainValue) to fetch previous range values. Because data is uploaded out of sequence (non-sequentially), this approach created a "visibility" bug:

Upload x=3: The query finds no previous data. It assigns y=1.1. No change is detected due to insufficient data.

Upload x=2: The query searches for v.data<=2. Since x=3 is the only record in the DB, the query returns null. The system assigns y=2.1 but fails to detect a change because it cannot "see" the x=3 record (since 3>2).

Upload x=1: The query searches for v.data<=1. Again, it returns null, failing to detect a change against x=2 or x=3.

The bug occurred because the v.data <= domain filter assumes that "previous" data always has a smaller domain value, which is false when uploads are unordered.

Then When I passed null for both domain node and domain value, it checks if(domain value !=null || domain node!null)
and then sorted using created_at. This regardless of the sequence of upload fetches the correct rangeValues based on the fingerprint value for that upload.

stalep:
Good catch — you're right that the `v.data <= domainValue` filter breaks for out-of-order uploads. When x=3 is uploaded first and then x=2 arrives, the query `v.data <= 2` can't see the x=3 record because 3 > 2, even though it was uploaded earlier and should be part of the comparison window.

The fallback to `created_at` ordering when both domain parameters are null makes sense for this case — it gives you the window based on upload history, which is what matters for detecting changes introduced by a new upload.

One thing worth adding: a test that explicitly validates the out-of-order upload scenario you described. Your current tests upload in descending domain order (x=3, x=2, x=1) which does exercise this path, but a test that uploads in a scrambled order (e.g., x=2, x=4, x=1, x=3) and verifies the correct changes are detected would document this behavior more explicitly and protect against future regressions.

src/test/java/io/hyperfoil/tools/h5m/svc/NodeServiceTest.java
```diff
@@ -1438,5 +1349,349 @@ public void calculateFixedThreshold_inclusive_vs_exclusive() throws SystemExcept
...
+
+        System.out.println("====Upload3======");
+        List<ValueEntity> changes3 = nodeService.calculateRelativeDifferenceValues(relDiff, root3, 0);
+
+        assertTrue(changes3.size() <= 1,
+            "Upload 3: Should process only x=1 (new domain value), producing at most 1 change");
+
+        ValueEntity change3 = changes3.get(0);
+        JsonNode changeData3 = change3.data;
+        double ratio3 = changeData2.get("ratio").asDouble();
```
stalep: Did you mean to use changeData3 here?

Arjun-Parmani: Yes. I'll rectify this!

stalep:
I think this looks good, @willr3 any objections?

src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
```diff
@@ -413,100 +413,133 @@ public List<ValueEntity> calculateRelativeDifferenceValues(RelativeDifference re
...
+                                null,
+                                (int) (relDiff.getWindow() + minPrevious),
+                                0,
+                                false
+                        );
+
+                        List<ValueEntity> allDomainValues = new ArrayList<>();
+                        allDomainValues.addAll(preceedingDomainValues);
+                        allDomainValues.addAll(followingDomainValues);
+                        for (int i = 0; i < allDomainValues.size(); i++) {
```
willr3: Did you come up with this method of de-duplication or did AI create this for you? I think this method is ignoring the fact that preceedingDomainValues is sorted and guaranteed to be <= uploadDomainValue and followingDomainValues is sorted and >= uploadDomainValue. This method is O(n^2) which is less efficient than the previous HashSet approach.
Do you see a problem with removing the initial values in followingDomainValues that == uploadDomainValue as we discussed? If you have an example where that would create an invalid change detection then please add that example to the test suite and we should use the HashSet approach you originally used.

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
```diff
@@ -999,4 +997,254 @@ public void fixedthreshold_qvss_split_by_framework(QuarkusMainLauncher launcher)
                 "quarkus-jvm should not appear in violations — scoping broken\n" + output);
     }
 
+    @Test
+    public void relativedifference_not_recalculate_old_changes(QuarkusMainLauncher launcher) throws IOException {
```
willr3, line 1001: I believe we talked about how this test is checking that the implementation of relative difference change detection fails to properly remove past detections. I do not believe we should add a test that is verifying the implementation is incomplete. This test would be removed or at the very least significantly changed when the relative difference detection correctly removes old change detections that are invalidated by new uploads. 
Please let me know if you feel new uploads should not remove past change detection notifications that now violate the minPrevoius. If that is the case then I agree we should keep this test.

src/test/java/io/hyperfoil/tools/h5m/svc/NodeServiceTest.java
```diff
@@ -1488,5 +1488,348 @@ public void calculateFixedThreshold_inclusive_vs_exclusive() throws SystemExcept
...
+        range1.sources = List.of(split1);
+        range1.persist();
+
+        ValueEntity fp1 = new ValueEntity(null, fingerprintNode, new TextNode("alpha"));
+        fp1.sources = List.of(split1);
+        fp1.persist();
+
+        tm.commit();
+
+        System.out.println("========Upload1======");
```
willr3: Please remove System.out.println from tests

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
```diff
@@ -999,4 +997,254 @@ public void fixedthreshold_qvss_split_by_framework(QuarkusMainLauncher launcher)
...
+                new String[]{"list","value","from",testName}
+        );
+
+        results.forEach(result->{
+            assertEquals(0,result.exitCode(),result.getOutput());
+        });
+
+        LaunchResult afterUpload2 = results.get(results.size() - 5);
+        String output2 = afterUpload2.getOutput();
+        System.out.println("After upload 2 (x=2):\n" + output2);
```
willr3: Please remove System.out.println from unit tests

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
```diff
@@ -999,4 +997,254 @@ public void fixedthreshold_qvss_split_by_framework(QuarkusMainLauncher launcher)
...
+        System.out.println("After upload 3 (x=3):\n" + output3);
+        assertTrue(output3.contains("Count: 17"),
+                "After upload 3, expect 17 values (2 changes total)\n" + output3);
+
+        assertTrue(output3.contains("\"domainvalue\":3") || output3.contains("domainvalue\": 3"),
+                "Change should be detected for domain x=3\n" + output3);
+
+        LaunchResult afterUpload4 = results.getLast();
+        String output4 = afterUpload4.getOutput();
+        System.out.println("After upload 4 (x=1):\n" + output4);
```
willr3: Please remove System.out.println from unit testing. If the information is important for failed asserts then please add it to the message for the assert.

src/test/java/io/hyperfoil/tools/h5m/svc/NodeServiceTest.java
```diff
@@ -1488,5 +1488,348 @@ public void calculateFixedThreshold_inclusive_vs_exclusive() throws SystemExcept
...
+        ValueEntity range2 = new ValueEntity(null, rangeNode, new DoubleNode(2.1));
+        range2.sources = List.of(split2);
+        range2.persist();
+
+        ValueEntity fp2 = new ValueEntity(null, fingerprintNode, new TextNode("alpha"));
+        fp2.sources = List.of(split2);
+        fp2.persist();
+        tm.commit();
+
+        System.out.println("====Upload2======");
```
willr3: I know I often forget to remove System.out.println but we should make a best effort to clean them up before pushing to the PR.

src/test/java/io/hyperfoil/tools/h5m/svc/NodeServiceTest.java
```diff
@@ -1488,5 +1488,348 @@ public void calculateFixedThreshold_inclusive_vs_exclusive() throws SystemExcept
...
+        List<ValueEntity> changes2 = nodeService.calculateRelativeDifferenceValues(relDiff, root2, 0);
+        assertEquals(1, changes2.size(),
+                "Upload 2: x=2 should produce 1 change (ratio exceeds 50% threshold)");
+
+        ValueEntity change2 = changes2.get(0);
+        JsonNode changeData2 = change2.data;
+        double ratio2 = changeData2.get("ratio").asDouble();
+        System.out.println("  Ratio: " + ratio2 + "%");
```
willr3: I'm going to stop highlighting all System.out.println but please try and remove them. If you feel they are adding important information to help diagnose a failing test then please add that information to the test's message.

src/test/java/io/hyperfoil/tools/h5m/svc/NodeServiceTest.java
```diff
@@ -1488,5 +1488,348 @@ public void calculateFixedThreshold_inclusive_vs_exclusive() throws SystemExcept
...
+
+        double previous3 = changeData3.get("previous").asDouble();
+        System.out.println("  Previous: " + previous3);
+
+        JsonNode domainValue3 = changeData3.get("domainvalue");
+        System.out.println("  domainvalue: " + domainValue3);
+
+        assertEquals(2, domainValue3.asInt(), "Domain should be 2");
+        assertEquals(3.1, previous3, "Previous range value should be 3.1");
+        //assertTrue(ratio3 > 20.0, "Ratio should exceed 20% threshold");
```
willr3: Why is the assert for the ratio commented out and based on such a large threshold?

src/test/java/io/hyperfoil/tools/h5m/svc/NodeServiceTest.java
```diff
@@ -1488,5 +1488,348 @@ public void calculateFixedThreshold_inclusive_vs_exclusive() throws SystemExcept
...
+
+        double previous3 = changeData3.get("previous").asDouble();
+        System.out.println("  Previous: " + previous3);
+
+        JsonNode domainValue3 = changeData3.get("domainvalue");
+        System.out.println("  domainvalue: " + domainValue3);
+
+        assertEquals(3, domainValue3.asInt(), "Domain should be 3");
+        assertEquals(2.1, previous3, 0.01, "Previous range value should be 2.1");
+        assertTrue(ratio3 > 20.0, "Ratio should exceed 20% threshold");
```
willr3: Why are you using assertTrue rather than assertEquals with a threshold?

willr3 (changes requested):
I think the overall change to calculate relative difference incrementally instead of re-scanning all of the persisted data is correct. I believe some of the code should be changed before we merge this PR. The de-duplication for domainValues appears inefficient. Please review that logic, remove the System.out.println, and help me understand why the ratio validation is relying on a boolean value comparison rather than verifying it calculated the correct value?

src/test/java/io/hyperfoil/tools/h5m/svc/NodeServiceTest.java
Arjun-Parmani: Yup. Removing all the System.out.println from unit test.

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
Arjun-Parmani: Yup. Removing all the System.out.println from unit case.

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
Arjun-Parmani: Removing all the System.out.println from unit case.

src/test/java/io/hyperfoil/tools/h5m/svc/NodeServiceTest.java
Arjun-Parmani: I think you are right. I should check the exact the ratio is produced. It currently just checks if it exceeds the threshold.

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
Arjun-Parmani: This test basically was written only to check and make sure that it is not scanning all of the Domain values based on the fingerprint like it did previously. It is not deleting any invalidated changes that occur after a new change is detected. I think what you mentioned about the removal of the past change detection is the significant next step. I think for now we can keep this test and later maybe use for testing the deletion of the past detections.

PR #81, opened 2026-05-03
Optimize sql/sqlall: batch jsonpath extractions to reduce DB round-trips

stalep:
## Summary
- Replaces INSERT-UPDATE-SELECT pattern with SELECT-first for sql/sqlall node value computation
- Batches all sibling sql/sqlall nodes sharing the same source into a single SQL query using PostgreSQL VALUES + CASE
- Adds SQLite batching support via individual queries with shared cache

## Problem
Each sql/sqlall node did 3-4 DB operations: INSERT value with null data, UPDATE with jsonpath result, SELECT to verify, DELETE if null. With ~90 sql/sqlall nodes per upload (rhivos-perf-comprehensive), that was ~270 DB round-trips.

## Fix
1. **SELECT-first**: Compute the jsonpath result directly via SELECT. Only INSERT values that have data. Null results skip all writes.
2. **Batch siblings (PostgreSQL)**: When the first sql/sqlall node for a source is processed, find all sibling nodes sharing that source and compute all jsonpaths in one query using `VALUES + CASE`. Cache results so subsequent siblings skip the DB entirely.
3. **Batch siblings (SQLite)**: Same sibling discovery and caching, but queries each sibling individually since SQLite lacks `jsonb_path_query_first`/`jsonb_path_query_array`. The first sibling triggers N individual queries, but the remaining N-1 siblings are served from cache — eliminating redundant source value lookups.

For rhivos on PostgreSQL (82 sql/sqlall nodes in 2 source groups), this reduces 82 individual queries to 2 batched queries. On SQLite, it reduces redundant DB round-trips by caching all sibling results on the first call.

## Benchmark — quarkus-spring-boot-comparison (100 uploads, PostgreSQL)

| Branch | Time | Values | vs main |
|--------|------|--------|---------|
| main | 47.1s | 4502 | baseline |
| load-legacy | 46.4s | 4502 | -1.5% (no regression) |
| **issue_80 (batched)** | **35.5s** | **4502** | **-24.5%** |

Same value count across all branches — no regressions. The load-legacy import changes have no impact on core upload performance.

## Benchmark — rhivos-perf-comprehensive (legacy import, 4 runs)

| | Baseline | Batched | Improvement |
|---|---------|-----------|-------------|
| Average | 113s | 103s | ~9% |

## Test plan
- [x] All 182 tests pass (SQLite — exercises the individual-query + cache path)
- [x] quarkus-spring-boot-comparison benchmark: same value count (4502), 24.5% faster
- [x] load-legacy branch: no regression (46.4s vs 47.1s, same values)
- [x] Verified rhivos import match rate unchanged (84.9%)

Fixes #80 (partial — addresses the sql/sqlall batching item)

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
@@ -595,7 +595,7 @@ public class NodeService implements NodeServiceInterface {
     }
     private List<ValueEntity> calculateSqlJsonpathValuesFirstOrAll(NodeEntity node, Map<String, ValueEntity> sourceValues, int startingOrdinal,String psqlFunction) throws IOException {
         List<ValueEntity> rtrn = new ArrayList<>();
-        if(sourceValues.isEmpty()){//end early when there isn't input
+        if(sourceValues.isEmpty()){
             return rtrn;
         }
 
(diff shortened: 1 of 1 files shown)


willr3:
This code cannot be merged in the current state.

1. There are 0 tests for the condition where the code works as expected.
2. There is no size bounds on `sqlBatchCache`
3. There is no timeout for entries in `sqlBatchCache`
4. There is no way to invalidate / evict entries from `sqlBatchCache`. For example,  if a `Node` changes and needs recalculation. This is just one example of the ramifications this PR would introduce and is by no means an exhaustive list of the conditions that would need to be tested.

I think this type of performance improvement was thought up because the current test suite runs a series of processes where each represents a cli operation. Whomever or whatever reviewed the performance did not pay any consideration to the process having to serve multiple requests or functions in parallel or the longevity of the process. 

The idea of performing sql jsonpath operations in batch does sound appealing but this implementation is going to cause more problems than in solves. Perhaps we should consider a `Work` that represents a batch operation and how we would handle `Work.activeNode` becoming a collection of `Nodes`?

willr3:
I created a `work_activeNodes` branch in my fork that is based on my fork of `load-legacy`. It changes `Work` to use a set of `activeNodes` instead of the single active node. I checked that all the existing tests work so it should be something you could base a "batch work" branch upon that would calculate all sql based values in one db operation similar to the original work in this PR but without introducing the `sqlBatchCache`. I have my fingers crossed this gives us another reduction in calculation time.

My fork of `load-legacy` was rebased onto `upstream/main` more recently than `upstream/load-legacy` and I play to update the `upstream/load-legacy` but won't do so until I can ask forgiveness for the `git push -f` it will require :)

PR #83, opened 2026-05-07
fix npe from calculateSourceValuePermutations

willr3:


willr3:
I thought I opened a separate PR for the fix JsNode getParameterNames but that one we need too :)

stalep:
The npe fix looks good, but remove the https://github.com/Hyperfoil/h5m/pull/83/commits/9bccf5b91ff9a274384a40e008a2075f048873eb commit from this pr first :)

PR #85, opened 2026-05-12
Optimize sql/sqlall: SELECT-first instead of INSERT-UPDATE-SELECT

stalep:
Instead of creating a value with null data, updating it via SQL, then reading it back to check for null — directly SELECT the jsonpath result, skip entirely if null, and only INSERT values that have data.

Before: INSERT + UPDATE + SELECT + (DELETE if null) = 3-4 DB ops per node
After:  SELECT + (INSERT if non-null) = 1-2 DB ops per node

For null results (missing jsonpath), eliminates all writes entirely.


willr3:
This was a humbling one. The initial testing showed the current `main` provided better performance but now that we have node graphs where more nodes have "children" than don't have children we see the reverse.

PR #86, opened 2026-05-12
Add indexes on value(node_id, folder_id) and remove redundant OR from…

stalep:
… recursive CTE base case

- Add @Table indexes for node_id and folder_id on ValueEntity, critical for CTE query performance when filtering results by node
- Remove 'OR ve.child_id = :rootId' from getDescendantValues and getDescendantValuesByNodes CTEs — the OR prevented efficient index usage on parent_id and was unnecessary (root is never its own descendant)
- Add tests verifying descendants exclude root when root shares the same node type as the query target


willr3:
Would you please change the commit message so that the first line highlights the fact that getDescendantValues was changed so that it no longer returns the root Value if the input `Node` is the same as `Value.node`?  That is probably that is a rather significant functional change that is buried on the last line.

PR #92, opened 2026-05-15
Add WebhookPlugin for HTTP POST notifications (issue #89, Phase 2)

stalep:
Implements the first notification channel: HTTP POST webhook. Sends a JSON payload containing folder name, detection node info, change details, and a formatted text message to a configured URL.

Features:
- Accepts URL via JSON config ({"url": "https://..."}) or plain URL
- Optional Authorization header via secrets ({"authHeader": "Bearer ..."})
- Custom message templates with {folderName}, {nodeName}, {nodeType}, {changeCount} placeholders
- Slack incoming webhook compatible (uses 'text' field in payload)
- Config validation: requires valid http/https URL
- 10s connect timeout, 30s request timeout
- Throws on HTTP 4xx/5xx responses for NotificationLog failure tracking

Tests (14):
- Validation: valid JSON config, plain URL, https, null, empty, non-http scheme, missing url field
- Send: JSON payload structure, auth header, custom template, default message, HTTP error handling, plain URL config
- Method identity check

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -174,6 +174,10 @@
         </dependency>
         <dependency>
             <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-qute</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
             <artifactId>quarkus-rest</artifactId>
         </dependency>
         <dependency>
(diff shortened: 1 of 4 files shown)


barreiro:
two questions about design choices:
- why not use vert.x web client ?? it should be a better option to do http requests from within quarkus
- could quarkus qute be used as the template engine ?? in here it's a simple replacement, but for consistency we should standardize on that

stalep:
> two questions about design choices:
> 
>     * why not use vert.x web client ?? it should be a better option to do http requests from within quarkus
> 
>     * could quarkus qute be used as the template engine ?? in here it's a simple replacement, but for consistency we should standardize on that

Good catch, fixed now!

PR #97, opened 2026-05-19
Web tests, dashboard and expression tester

stalep:


src/main/java/io/hyperfoil/tools/h5m/rest/FolderResource.java
```diff
@@ -26,13 +29,24 @@ public class FolderResource {
     @Inject
     FolderServiceInterface folderService;
 
+    @Inject
+    FolderService folderServiceImpl;
```
willr3: What is the reason to inject both the API and the impl?

src/main/java/io/hyperfoil/tools/h5m/svc/FolderService.java
```diff
@@ -91,6 +104,86 @@ public List<Folder> list(){
...
+                (String) row[1],
+                ((Number) row[2]).intValue(),
+                ((Number) row[3]).intValue(),
+                ((Number) row[4]).intValue(),
+                toLocalDateTime(row[5]),
+                toLocalDateTime(row[6])
+            ));
+        }
+        return summaries;
+    }
```
willr3, line 147: That is a scary number of joins. Are we expecting this to be an entity that is cached to avoid so many full table scans?

src/main/java/io/hyperfoil/tools/h5m/api/UploadSummary.java
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Summary of an upload (root value) for listing in the folder detail view.
+ */
+@Schema(description = "Summary of an uploaded data entry")
+public record UploadSummary(
+    @Schema(description = "Value ID") long id,
+    @Schema(description = "Upload timestamp") LocalDateTime createdAt,
+    @Schema(description = "Number of computed descendant values") int valueCount
+) {}
```
willr3: I don't userstand the use case for this entity. The upload either has the desired values or it doesn't and that will likely be apparent in whatever view the user creates to visualize their uploads or in their change detection. I cannot think of a user case where they need a summary of the individual uploads. What was the user use case for this entity?

willr3 (commented):
I think the UploadSummary may not be worth it and the FolderSummary sql is cause for concern. Did FolderSummary appear to scale well?

src/main/java/io/hyperfoil/tools/h5m/rest/FolderResource.java
stalep: The `getDashboardSummaries()` method is not exposed on the interface. I can change that.

src/main/java/io/hyperfoil/tools/h5m/svc/FolderService.java
stalep: The number of joins is quite a bit, but it looks worse than it is I think since several are left joins for optional counts...

src/main/java/io/hyperfoil/tools/h5m/api/UploadSummary.java
stalep: Hm, the idea was to show a paginated list of uploads, but with the views feature now planned it could be removed/replaced by the view data table. I'll remove it.

PR #102, opened 2026-05-22
add load-legacy-tests, load-legacy-runs, and verifyimport commands

willr3:
* add sqlall to jq migration for legacy extractors
* add PostgreSQL jsonpath filter expression conversion to JQ
* only queue top level nodes for upload
* change Work to a set of activeNodes

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/cli/LoadLegacyTests.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/cli/LoadLegacyTests.java
@@ -1,0 +1,892 @@
+package io.hyperfoil.tools.h5m.cli;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import io.agroal.api.AgroalDataSource;
+import io.agroal.api.configuration.supplier.AgroalPropertiesReader;
+import io.hyperfoil.tools.h5m.entity.FolderEntity;
+import io.hyperfoil.tools.h5m.entity.NodeEntity;
+import io.hyperfoil.tools.h5m.entity.NodeGroupEntity;
+import io.hyperfoil.tools.h5m.entity.node.*;
+import io.hyperfoil.tools.h5m.svc.FolderService;
+import io.hyperfoil.tools.h5m.svc.NodeService;
+import io.hyperfoil.tools.yaup.HashedLists;
+import io.hyperfoil.tools.yaup.HashedSets;
+import io.hyperfoil.tools.yaup.StringUtil;
+import jakarta.inject.Inject;
+import picocli.CommandLine;
+
+import java.sql.*;
+import java.util.*;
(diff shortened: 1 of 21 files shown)


src/main/java/io/hyperfoil/tools/h5m/cli/LoadLegacyTests.java
```diff
@@ -0,0 +1,892 @@
...
+                    return false;
+                }
+                if(extractors.size()!=t.extractors.size()){
+                    return false;
+                }
+                for(int i=0;i<extractors.size();i++){
+                    if(!extractors.get(i).equals(t.extractors.get(i))){
+                        return false;
+                    }
+                }
```
stalep, line 187: Should there be a `return true;` here?
This method will always return false atm.

src/main/java/io/hyperfoil/tools/h5m/entity/NodeEntity.java
```diff
@@ -38,7 +38,7 @@ public static enum ScalarVariableMethod { First, All}
 
     //making this eager causes too many joins
     //cannot cascade delete because this entity "owns" the reference to the parent values
-    @ManyToMany(cascade = { CascadeType.PERSIST, CascadeType.MERGE }, fetch = FetchType.LAZY )
+    @ManyToMany(cascade = { CascadeType.PERSIST , CascadeType.REFRESH, CascadeType.MERGE, CascadeType.DETACH }, fetch = FetchType.LAZY )
```
stalep: I'm thinking we also only need `CascadeType.PERSIST` here, neither refresh, merge or detatch is needed (I think).

src/main/java/io/hyperfoil/tools/h5m/entity/NodeEntity.java
```diff
@@ -29,7 +29,7 @@ public static enum ScalarVariableMethod { First, All}
     public MultiIterationType multiType = MultiIterationType.Length;
     public ScalarVariableMethod scalarMethod = ScalarVariableMethod.First;
 
-    @ManyToOne(cascade = { CascadeType.PERSIST, CascadeType.MERGE }, fetch = FetchType.LAZY )
+    @ManyToOne(cascade = { CascadeType.PERSIST, CascadeType.REFRESH }, fetch = FetchType.LAZY )
```
stalep: I'm thinking we only need CascadeType.PERSIST here?

src/main/java/io/hyperfoil/tools/h5m/cli/LoadLegacyTests.java
willr3: The `return true;` should have been at the end :)

src/main/java/io/hyperfoil/tools/h5m/entity/NodeEntity.java
willr3: I revered to what I found in `main` so we can introduce changes in a subsequent PR. maintaining the scope of change to a minimum to help us whenever we look at the commits in the future :)

PR #106, opened 2026-05-26
Removing invalid changes in relative difference calculation after new…

Arjun-Parmani:
***Draft PR*** 
After a change is detected, I am fetching a persisted change and checking if that domain values is within the allDomainValues and then deleting the invalid change

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/entity/NodeEntity.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/entity/NodeEntity.java
@@ -93,13 +93,6 @@ public abstract class NodeEntity extends PanacheEntity implements Comparable<Nod
         this.previousVersion = previousVersion;
     }
 
-    @PreUpdate
-    @PrePersist
-    public void sortSources(){
-        this.sources = KahnDagSort.sort(sources, NodeEntity::getSources);
-    }
-
-
     public NodeEntity(){
         this.sources = new ArrayList<>();
     }
--- a/src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
@@ -41,7 +41,6 @@ import io.hyperfoil.tools.jjq.value.JqArray;
 import io.hyperfoil.tools.jjq.value.JqValue;
 
 import java.io.IOException;
-import java.sql.PreparedStatement;
 import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.DoubleBinaryOperator;
@@ -438,11 +437,13 @@ public class NodeService implements NodeServiceInterface {
                         //this would need to filter rangeValues that already have a changeDetection detected...
                     }
                     */
+
                     //changing domainValues to just get the domainValues from this root would change from full series scanning to just scanning the new values
                     //but that would only work if values are added sequentially to the domain value (or we delay relative difference calculation to the end of the work queue.
                     //perhaps we check if root introduced the maximum domainValue then only calculate new changes for that last window
                     //or get the domainValues greater than domain values from root and calculate all those changes?
                     List<ValueEntity> rootDomainValues = valueService.getDescendantValues(root, relDiff.getDomainNode());
+                    List<ValueEntity> allDomainValues = new ArrayList<>();
                     for (int sdIdx = 0; sdIdx < rootDomainValues.size(); sdIdx++) {
                         ValueEntity uploadedDomainValue = rootDomainValues.get(sdIdx);
                         List<ValueEntity> preceedingDomainValues = valueService.findMatchingFingerprint(
@@ -468,12 +469,11 @@ public class NodeService implements NodeServiceInterface {
                                 0,
                                 false
                         );
-                        followingDomainValues.remove(0);
-
-                        List<ValueEntity> allDomainValues = new ArrayList<>();
+                        if(!followingDomainValues.isEmpty()) {
+                            followingDomainValues.remove(0);
+                        }
                         allDomainValues.addAll(preceedingDomainValues);
                         allDomainValues.addAll(followingDomainValues);
-
                         for (int dIdx = 0; dIdx < allDomainValues.size(); dIdx++) {
                             ValueEntity domainValue = allDomainValues.get(dIdx);
                             //todo this does not look for values after previous relDiff observation :(
@@ -553,13 +553,52 @@ public class NodeService implements NodeServiceInterface {
                                     dIdx += minPrevious;
                                     ValueEntity changeValue = new ValueEntity(root.folder, relDiff, data);
                                     changeValue.idx = startingOrdinal;
-                                    List<ValueEntity> foundParents = valueService.getAncestor(fingerprintValue, groupBy);
+                                    List<ValueEntity> foundParents = valueService.getAncestor(domainValue, groupBy);
                                     if (foundParents.size() == 1) {
                                         changeValue.sources = foundParents;
                                     }
+
                                     rtrn.add(changeValue);
                                 }
                             }
+                        }
+                        List<ValueEntity> persistedChangeValues = valueService.findMatchingFingerprint(
+                                relDiff,
+                                groupBy,
+                                fingerprintValue,
+                                relDiff.getDomainNode(),
+                                uploadedDomainValue,
+                                null,
+                                (int) (relDiff.getWindow() + minPrevious),
+                                0,
+                                false
+                        );
+                        List<JsonNode> domainRemoveScope = new ArrayList<>();
+                        for (ValueEntity dv : allDomainValues) {
+                            if (dv.data != null) {
+                                domainRemoveScope.add(dv.data);
+                            }
+                        }
+                        if (!rtrn.isEmpty()) {
+                            for (ValueEntity existingValue : persistedChangeValues) {
+                                JsonNode existingDomainValue = existingValue.data.get("domainvalue");
+                                if (existingDomainValue == null) continue;
+                                if (domainRemoveScope.contains(existingDomainValue)) {
+                                    boolean match = false;
+                                    for (ValueEntity currentValue : rtrn) {
+                                        JsonNode currentDomainValue = currentValue.data.get("domainvalue");
+                                        if (currentDomainValue == null) continue;
+                                        if (existingDomainValue.equals(currentDomainValue)) {
+                                            match = true;
+                                            break;
+                                        }
+                                    }
+                                    if (!match) {
+                                        valueService.delete(existingValue);
+                                    }
+                                }
+                            }
+
                         }
                     }
                 }
--- a/src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
+++ b/src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
@@ -1055,7 +1055,7 @@ public class H5mTest {
                 }
                 """
         );
-        
+
         List<LaunchResult> results = run(launcher,
                 new String[]{"add","folder",testName},
                 new String[]{"add","jq","to",testName,"split",".Item[]"},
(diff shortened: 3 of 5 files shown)


src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
```diff
@@ -567,6 +540,36 @@ public List<ValueEntity> calculateRelativeDifferenceValues(RelativeDifference re
         }catch (Exception e){
             e.printStackTrace();
         }
+
+        List<ValueEntity> existingValues = valueService.getValues(relDiff);
```
willr3: I believe there's a bug here. Can you create a test that would fail because of this line?

src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
```diff
@@ -567,6 +540,36 @@ public List<ValueEntity> calculateRelativeDifferenceValues(RelativeDifference re
...
+        List<Long> DomainRemoveScope = new ArrayList<>();
```
willr3: This is not idiomatic Java. Java uses capital letters at the start of a class name not variables.

src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
```diff
@@ -567,6 +540,36 @@ public List<ValueEntity> calculateRelativeDifferenceValues(RelativeDifference re
...
+        for (ValueEntity dv : allDomainValues) {
+            if (dv.data != null && dv.data.isNumber()) {
+                DomainRemoveScope.add(dv.data.asLong());
```
willr3: I do not believe domainValues are confined to data that is a long. It can be anything that is sortable / comparable. Perhaps change one of the tests to use Double, or String so that you can ensure the code does not assume Long?

src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
```diff
@@ -567,6 +540,36 @@ public List<ValueEntity> calculateRelativeDifferenceValues(RelativeDifference re
...
+                   for (ValueEntity currentValue : rtrn) {
+                       JsonNode currentDomainValue = currentValue.data.get("domainvalue");
+                       if (currentDomainValue == null || !currentDomainValue.isNumber()) continue;
+                       if (existingdomainValue.equals(currentDomainValue)) {
+                           match = true;
+                           break;
+                       }
+                   }
+                   if (!match) {
+                       existingValue.delete();
```
willr3: This should call a method on valueService to delete the entity. Otherwise we violate the abstraction that the ValueService manages the Values.

src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
```diff
@@ -567,6 +540,36 @@ public List<ValueEntity> calculateRelativeDifferenceValues(RelativeDifference re
...
+        List<Long> DomainRemoveScope = new ArrayList<>();
+        for (ValueEntity dv : allDomainValues) {
+            if (dv.data != null && dv.data.isNumber()) {
+                DomainRemoveScope.add(dv.data.asLong());
+            }
+        }
+       if(!rtrn.isEmpty()){
+           for(ValueEntity existingValue : existingValues){
+               JsonNode existingdomainValue = existingValue.data.get("domainvalue");
+               if (existingdomainValue == null || !existingdomainValue.isNumber()) continue;
```
willr3: I mentioned earlier but relative difference needs to work for any sortable domain data.

willr3 (changes requested):
There are some idiomatic and functional issues in the PR. Please review and let me know if you have any questions.

src/main/java/io/hyperfoil/tools/h5m/entity/NodeEntity.java
```diff
@@ -93,11 +93,11 @@ public void setPreviousVersion(NodeEntity previousVersion) {
         this.previousVersion = previousVersion;
     }
 
-    @PreUpdate
+    /*@PreUpdate
     @PrePersist
     public void sortSources(){
         this.sources = KahnDagSort.sort(sources, NodeEntity::getSources);
-    }
+    }*/
```
stalep: Better to remove this than have it commented out if it's not needed

src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
```diff
@@ -438,11 +437,13 @@ public List<ValueEntity> calculateRelativeDifferenceValues(RelativeDifference re
                         //this would need to filter rangeValues that already have a changeDetection detected...
                     }
                     */
+
                     //changing domainValues to just get the domainValues from this root would change from full series scanning to just scanning the new values
                     //but that would only work if values are added sequentially to the domain value (or we delay relative difference calculation to the end of the work queue.
                     //perhaps we check if root introduced the maximum domainValue then only calculate new changes for that last window
                     //or get the domainValues greater than domain values from root and calculate all those changes?
                     List<ValueEntity> rootDomainValues = valueService.getDescendantValues(root, relDiff.getDomainNode());
+                    List<ValueEntity> allDomainValues = new ArrayList<>();
```
stalep, line 446: What happens with `allDomainValues` if there is an upload with multiple root domain values?
Is it intentional that `allDomainValues` are not reset?

src/test/java/io/hyperfoil/tools/h5m/svc/NodeServiceTest.java
```diff
@@ -1988,4 +1990,470 @@ public void testCalculateRelativeDifferenceValues_unordered_sequence() throws Ex
...
+        ValueEntity split1 = new ValueEntity(null, splitNode, mapper.createObjectNode());
+        split1.sources = List.of(root1);
+        split1.persist();
+
+        ValueEntity domain1 = new ValueEntity(null, domainNode, new LongNode(4));
+        domain1.sources = List.of(split1);
+        domain1.persist();
+
+        ValueEntity range1 = new ValueEntity(null, rangeNode, new DoubleNode(400));
+        range1.sources = List.of(split1);
```
stalep, line 2176: I do not think range1.sources should be sourced from split1?

src/test/java/io/hyperfoil/tools/h5m/svc/ValueServiceTest.java
```diff
@@ -790,6 +790,66 @@ public void findMatchingFingerprint_sorting_cousin() throws SystemException, Not
 
     }
     @Test
+    public void findMatchingFingeprint_dataset_sibliling() throws SystemException, NotSupportedException, HeuristicRollbackException, HeuristicMixedException, RollbackException, JsonProcessingException {
```
stalep: spelling in method name

stalep (commented):
I do not think we should manually create any ValueEntites in tests... - I think we should try to use the services to upload json data such that we use the h5m pipeline and it will create the value DAG correctly (hopefully).
I think that's less error prone and also makes the tests better as it use a larger part of the codebase.

src/test/java/io/hyperfoil/tools/h5m/svc/NodeServiceTest.java
Arjun-Parmani: The structure of the nodes created needs to be rectified. Should be same as est_calculaterelativediff_remove_invalid_change_detection_following_domainValue(). Where Split node is based off of root node and domain,range,fingerprint are based off of the split node. After rectifying that i don't think range1.sources(split1) should be changed. Let me know your thoughts.

src/test/java/io/hyperfoil/tools/h5m/svc/ValueServiceTest.java
Arjun-Parmani: I'll rectify this

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
```diff
@@ -1249,16 +1249,16 @@ public void relativedifference_Unordered_uploads(QuarkusMainLauncher launcher) t
 
         LaunchResult afterUpload3 = results.get(results.size() - 3);
         String output3 = afterUpload3.getOutput();
-        assertTrue(output3.contains("Count: 17"),
-                "After upload 3, expect 17 values (2 changes total)\n" + output3);
+        assertTrue(output3.contains("Count: 16"),
+                "After upload 3, expect 16 values (2 changes total)\n" + output3);
```
willr3: Relying on the total number of values in the view was a quick way to write early tests but we should be replacing those asserts with asserts about the content of values not just the total count. Also is the match still correct where there are 2 changes when the number of values decreases to 16?

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
```diff
@@ -1249,16 +1249,16 @@ public void relativedifference_Unordered_uploads(QuarkusMainLauncher launcher) t
...
 
         assertTrue(output3.contains("\"domainvalue\":3") || output3.contains("domainvalue\": 3"),
                 "Change should be detected for domain x=3\n" + output3);
 
         LaunchResult afterUpload4 = results.getLast();
         String output4 = afterUpload4.getOutput();
-        assertTrue(output4.contains("Count: 23"),
-                "After upload 4, expect 23 values (3 changes total)\n" + output4);
+        assertTrue(output4.contains("Count: 22"),
+                "After upload 4, expect 22 values (3 changes total)\n" + output4);
```
willr3: The total number of values decreased by 1 but the expected number of changes is still 3? Can we change this assert to something that verifies the number of changes rather than the total number of values?

src/test/java/io/hyperfoil/tools/h5m/svc/NodeServiceTest.java
```diff
@@ -1988,4 +1990,470 @@ public void testCalculateRelativeDifferenceValues_unordered_sequence() throws Ex
...
+
+        ValueEntity range1 = new ValueEntity(null,rangeNode,split1.data.get("range"),List.of(split1));
+        range1.persist();
+
+        ValueEntity fingerprint1 = new ValueEntity(null,fingerprintNode,split1.data.get("fingerprint"),List.of(split1));
+        fingerprint1.persist();
+
+        tm.commit();
+
+        List<ValueEntity> changes1 = nodeService.calculateRelativeDifferenceValues(relDiff, root1, 0);
```
willr3, line 2039: Should we assert that there are 0 changes detected?

src/test/java/io/hyperfoil/tools/h5m/svc/NodeServiceTest.java
```diff
@@ -1988,4 +1990,470 @@ public void testCalculateRelativeDifferenceValues_unordered_sequence() throws Ex
...
+        NodeEntity rootNode = new RootNode();
+        rootNode.persist();
+        NodeEntity rangeNode = new JqNode("range",".y",rootNode);
+        rangeNode.persist();
+        NodeEntity domainNode = new JqNode("domain",".domain",rootNode);
+        domainNode.persist();
+        NodeEntity fingerprintNode = new JqNode("fingerprint",".fingerprint",rootNode);
+        fingerprintNode.persist();
+
+        SplitNode splitNode = new SplitNode("split", "split", List.of(rootNode));
```
willr3: are `range`, `domain`, and `fingerprint` meant to extract from `split`?

willr3 (changes requested):

src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
Arjun-Parmani: Initially the scope of the clean up code was written outside of the Root domain value. So List<ValueEntity> allDomainValues was then initialized outside of this scope so that it can be fetched in the clean up code

src/test/java/io/hyperfoil/tools/h5m/svc/NodeServiceTest.java
Arjun-Parmani: added a check to check 0 changes are detected

PR #112, opened 2026-05-29
Add Default view auto-creation, Data tab, and view tests (issue #105,…

stalep:
… Phase 3-5a)

Phase 3: Default view auto-creation
- Auto-create 'Default' view on folder create() and importFolder()
- Leaf nodes only (no detection nodes ft/rd/ed)
- Duplicate node names get suffix for unique constraint
- FolderService.delete() cleans up views before folder deletion

Phase 4: Tests
- Default view created for new folder (empty) and on rhivos import (with leaf nodes)
- View data with multiple uploads returns one row per upload
- View component ordering verified
- Updated existing view tests for Default view presence

Phase 5a: Data tab
- New DataTab component with Carbon Dropdown view selector and DataTable
- Columns from view components ordered by headerOrder
- Prefers views with components when defaulting (skips empty Default)
- Data tab is the first tab in FolderPage
- Updated FolderPage tests for three tabs (Data, Nodes, Graph)
- Updated design doc with Phase 5 details

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/svc/FolderService.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/svc/FolderService.java
@@ -12,6 +12,7 @@ import io.hyperfoil.tools.h5m.entity.NodeEntity;
 import io.hyperfoil.tools.h5m.entity.NodeGroupEntity;
 import io.hyperfoil.tools.h5m.entity.Team;
 import io.hyperfoil.tools.h5m.entity.ValueEntity;
+import io.hyperfoil.tools.h5m.entity.ViewEntity;
 import io.hyperfoil.tools.h5m.entity.mapper.ApiMapper;
 import io.hyperfoil.tools.h5m.entity.node.*;
 import io.hyperfoil.tools.h5m.entity.work.Work;
@@ -27,10 +28,7 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.time.LocalDateTime;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 @ApplicationScoped
 public class FolderService implements FolderServiceInterface {
@@ -67,6 +65,7 @@ public class FolderService implements FolderServiceInterface {
         entity.name = name;
         entity.group = new NodeGroupEntity(name); //TODO do we auto-create a nodeGroup?
         FolderEntity.persist(entity);
+        createDefaultView(entity);
         return entity.id;
     }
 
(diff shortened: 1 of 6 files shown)


src/main/java/io/hyperfoil/tools/h5m/svc/FolderService.java
```diff
@@ -401,10 +406,68 @@ public String importFolder(Path inputPath, boolean overwrite) throws IOException
...
+
         Log.infof("Imported folder '%s' with %d nodes from %s", folderName, nodeArray.size(), inputPath);
         return folderName;
     }
 
+    /**
+     * Creates or replaces a "Default" view for the folder containing leaf nodes —
+     * nodes that no other node depends on, excluding change detection nodes (ft, rd, ed).
+     */
+    private void createDefaultView(FolderEntity folder) {
```
willr3, line 417: I do not think a createDefaultView method should be able to delete the existing View. It looks like the goal of this method is to ensure the View exists and has a ViewComponent for each "leaf" NodeEntity in the NodeGroup. Perhaps it would be better if this merged the existing View's components with newly created ones for the "leaf" NodeEntities?

Why was deleting the existing view part of the scope of this method?

src/main/java/io/hyperfoil/tools/h5m/svc/FolderService.java
```diff
@@ -401,10 +406,68 @@ public String importFolder(Path inputPath, boolean overwrite) throws IOException
...
+        Set<Long> referencedAsSource = new HashSet<>();
+        for (NodeEntity node : folder.group.sources) {
+            if (node.sources != null) {
+                for (NodeEntity src : node.sources) {
+                    if (src.id != null) {
+                        referencedAsSource.add(src.id);
+                    }
+                }
+            }
+        }
```
willr3: This is going to load all nodes in the group, would it be better to use a db query to find the nodes that are leafs?

willr3 (changes requested):

src/main/java/io/hyperfoil/tools/h5m/svc/FolderService.java
stalep: Yeah, that was not supposed to be there. I played with creating it in importFolder() and create(). That was not good. Fixed now!

src/main/java/io/hyperfoil/tools/h5m/svc/FolderService.java
stalep: Yeah, I do not know what's best tbh. It was not easy to get right. For one project it might work to use all the leaf nodes, but for rhivos it just falls apart.
I've just removed auto-geneating any views now. atm it's empty.

PR #118, opened 2026-06-03
Transient Work with CompletableFuture and crash recovery tracking

stalep:
## Summary

Remove Work persistence from the hot path — Work items exist only in memory, eliminating orphaned DB rows from concurrent worker duplicate rejection. Upload returns a `CompletableFuture<Void>` for per-upload completion tracking.

## Key Changes

- **WorkService.create()**: No `em.merge/em.flush` — Work is never persisted to the DB. Only pre-computes ancestor caches for work queue sort.
- **WorkService.execute()**: Uses source values directly from the Work object (data already in memory) instead of reloading from DB. Uses `Hibernate.isPropertyInitialized()` to detect if data needs loading (fallback for tests calling execute() directly). Active nodes are still reloaded for lazy `sources` access.
- **UploadTracker**: `AtomicInteger` counter + `CompletableFuture` per upload, handles success, failure, early return, and rejected duplicates
- **FolderService.upload()**: Returns `CompletableFuture<Void>`, uses `QuarkusTransaction.requiringNew()` to avoid `@Transactional` + `CompletionStage` deadlock
- **UploadProcessingEntity**: Lightweight crash recovery — tracks whether each upload completed. On startup, incomplete uploads are re-triggered.
- **LoadLegacyRuns**: Batched uploads with `CompletableFuture.allOf()`

## Benchmark: 20 rhivos runs, core=4

| Metric | Main | This branch |
|--------|------|-------------|
| **Import time** | **330s** | **308s (-7%)** |
| Work items left in DB | 27 orphans | 0 |
| Values | 8,700 | 8,700 |
| Value edges | 9,740 | 9,740 |
| Tracking records | N/A | 20/20 completed |

The 7% speedup comes from keeping source value JSON data in memory through the work queue instead of round-tripping through PostgreSQL JSONB (serialize → store → load → deserialize). CPU profiling showed JSON deserialization (`DatabaseJsonFormatMapper.fromString`) was the #1 hotspot at ~2000 samples.

## Optimization Progression

| Step | Time | vs Main |
|------|------|---------|
| Initial (sequential .join()) | 381s | +15% |
| Batched `allOf()` | 358s | +8% |
| Remove unnecessary `Hibernate.initialize()` | 345s | +5% |
| Skip source value DB reload | **308s** | **-7%** |

## Trade-offs

**Positives:**
- 7% faster than main for rhivos import
- Zero orphaned Work rows (vs 27 on main with core=4)
- Per-upload `CompletableFuture` — no more polling `isIdle()` with Thread.sleep
- Fewer DB operations per work item (no persist/find/delete cycle, no source value reload)
- Crash recovery via `UploadProcessingEntity`

**Negatives:**
- No per-work-item crash recovery (entire upload is re-processed)
- `QuarkusTransaction.requiringNew()` instead of `@Transactional` on `upload()` is less obvious
- Source values held in memory longer (through the work queue lifetime)

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/entity/work/Work.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/entity/work/Work.java
@@ -4,53 +4,25 @@ import io.hyperfoil.tools.h5m.entity.NodeEntity;
 import io.hyperfoil.tools.h5m.entity.ValueEntity;
 import io.hyperfoil.tools.h5m.entity.node.RelativeDifference;
 import io.hyperfoil.tools.h5m.svc.WorkService;
-import io.quarkus.hibernate.orm.panache.PanacheEntity;
 import jakarta.enterprise.inject.spi.CDI;
-import jakarta.persistence.*;
-import org.hibernate.annotations.BatchSize;
 
 import java.util.*;
 import java.util.stream.Collectors;
 
 
-@Entity(name = "work")
-@DiscriminatorColumn(name = "type")
-@DiscriminatorValue("node")
 //cross test comparison could use sourceNodes and not have an activeNode?
 //custom post nodegroup actions could have sourceNodes without activeNode
-public class Work  extends PanacheEntity implements Runnable, Comparable<Work>{
+public class Work implements Runnable, Comparable<Work>{
 
-    @BatchSize(size=10)
-    @ManyToMany(cascade = {CascadeType.PERSIST}, fetch = FetchType.LAZY)
-    @JoinTable(
-            name="work_values",
-            joinColumns = @JoinColumn(name = "work_id"),
-            inverseJoinColumns = @JoinColumn(name = "value_id")
-    )
     public List<ValueEntity> sourceValues;//multiple values could happen for cross test comparisons and
 
-    @BatchSize(size=10)
-    @ManyToMany(cascade = {CascadeType.PERSIST}, fetch = FetchType.LAZY)
-    @JoinTable(
-            name="work_source_nodes",
-            joinColumns = @JoinColumn(name = "work_id"),
-            inverseJoinColumns = @JoinColumn(name = "node_id")
-    )
     public List<NodeEntity> sourceNodes; //what is going to use a list of sources that are not already listed for the activeNode?
 
     public int retryCount;
 
-    @BatchSize(size=10)
-    @ManyToMany(cascade = {CascadeType.PERSIST}, fetch = FetchType.EAGER)
-    @JoinTable(
-            name="work_active_nodes",
-            joinColumns = @JoinColumn(name = "work_id"),
-            inverseJoinColumns = @JoinColumn(name = "node_id")
-    )
     public Set<NodeEntity> activeNodes;
 
-    boolean cumulative = false;
-
+    public boolean cumulative = false;
 
     public Work(){
         retryCount = 0;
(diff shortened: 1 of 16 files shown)


src/main/java/io/hyperfoil/tools/h5m/entity/work/Work.java
```diff
@@ -49,7 +49,11 @@ public class Work  extends PanacheEntity implements Runnable, Comparable<Work>{
     )
     public Set<NodeEntity> activeNodes;
 
-    boolean cumulative = false;
+    public boolean cumulative = false;
+
+    /** Root value ID for tracking upload completion. Not persisted. */
+    @Transient
+    public long rootValueId;
```
willr3: Work can have multiple `sourceValues` which is why we used a list. Right now all the calculations for uploads have a single sourceValue but users wanted cross upload and cross test calculations so we designed `Work` to be able to accept multiple source values.

src/main/java/io/hyperfoil/tools/h5m/svc/WorkService.java
```diff
@@ -127,27 +134,62 @@ public void beforeCompletion() {}
...
+    /**
+     * Creates work items with upload completion tracking.
+     * Returns a CompletableFuture that completes when all work
+     * (including cascaded child work) finishes.
+     */
     @Transactional
-    public void delete(Work work){
-        if(work.id!=null){
-            Work.deleteById(work.id);
+    public CompletableFuture<Void> createTracked(List<Work> works, long rootValueId) {
```
willr3: Do you think it would work to have this method iterate over each `Work` in `works` then each `sourceValues` entry to update the tracker? This would allow the trackers to use the existing `sourceValues` and to work when a `Work` has more than one upload in the `sourceValues`.

stalep:
This change should have feature parity with main in addition to supporting CompleteableFuture and imo a simpler recovery model.

src/main/java/io/hyperfoil/tools/h5m/rest/FolderResource.java
```diff
@@ -77,12 +77,15 @@ public long delete(@PathParam("name") String name) {
     @POST
     @Path("{name}/upload")
     @Authenticated
-    @Operation(description = "Upload JSON data to a folder")
+    @Operation(description = "Upload JSON data to a folder and wait for processing to complete")
```
willr3: If this is now a synchronous call then does that mean folderService.upload is now synchronous? I thought it was async and returned a `CompletableFuture`?

src/main/java/io/hyperfoil/tools/h5m/svc/WorkService.java
```diff
@@ -63,15 +66,14 @@ public class WorkService implements WorkServiceInterface {
 
     private WorkQueueExecutor workExecutor;
 
+    private final ConcurrentHashMap<Long, UploadTracker> trackers = new ConcurrentHashMap<>();
+
     @Transactional
     void onStart(@Observes StartupEvent ev) {
```
willr3: Is this in a race with the `FolderService.recoverIncompleteUploads` or is there something that guarantees the `WorkService` is started before the `FolderService` ?

src/test/java/io/hyperfoil/tools/h5m/rest/RestEndpointTest.java
```diff
@@ -843,6 +829,57 @@ public void view_component_ordering() throws Exception {
                 .body("components[1].headerOrder", equalTo(1));
     }
 
+    // -- Crash recovery --
+
+    @Test
+    public void incomplete_upload_is_reprocessed_on_recovery() throws Exception {
```
willr3: Why is this on the RestEndpointTest when it is testing the functionality of the folderService? I think we need a Unit tests on `FolderServiceTest` that verify `recoverIncompleteUploads` functions correctly given the possible input states. This looks like an integration test that verifies the components work together but does not cover the use cases for `recoverIncompleteUploads`

willr3 (changes requested):
I'm excited for this change but I think there are a few details to resolve before we merge.

PR #124, opened 2026-06-10
Added a new Rest end point for label Values in h5m

Arjun-Parmani:
**Description of the PR**
  
- Added getLabelValues API endpoint (GET /api/folder/{name}/labelValues) that returns one JSON row per upload for a folder, with each row containing values keyed by node name
- Supports optional query params: nodes (filter which nodes appear as columns), groupbyId (include grouping node in every row), sortById (sort rows by a node's value ascending)
- Refactored getGroupedValues(Long, List<Long>) to delegate to the new 3-param overload getGroupedValues(Long, List<Long>, Long) eliminating duplicate SQL
- Added NotFoundException when folder is not found (returns 404)

  **Test plan**
  
ValueServiceTest:  
- getGroupedValues_filtered — verifies that passing filterNodeIds excludes nodes not in the list from the result
- getGroupedValues_sorted — verifies that passing sortByNodeId returns rows in ascending order by that node's value,   regardless of insertion order
- getGroupedValues_empty_filterNodeIds_returns_all_nodes — verifies that passing an empty list falls back to returning all nodes
- getLabelValues_empty_filterNodeIds_returns_all_nodes — verifies the service-level method returns all nodes and all uploads when no filter is applied

RestEndpointTest: 
- labelValues_returns_grouped_values -REST level integration test covering unfiltered fetch and  sorted fetch via HTTP
- labelValues_folder_not_found_returns_404 — verifies 404 is returned for a non-existent folder

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/api/svc/ValueServiceInterface.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/api/svc/ValueServiceInterface.java
@@ -48,4 +48,17 @@ public interface ValueServiceInterface {
      */
     List<Value> getNodeValues(Long nodeId);
 
+    /**
+         * Returns one row per upload for a folder, with each row containing the values of the requested nodes.
+         * groupByNodeId and sortByNodeId are always included in every row regardless of nodeIds.
+         * @param folderId Folder Id to query.
+         * @param nodeIds Node IDs to include as columns (null/empty list = all nodes).
+         * @param groupByNodeId Node whose value identifies the series (e.g. config fingerprint).
+         * @param sortByNodeId  Node whose value orders the rows (acts as X-axis).
+       */
+      List<JsonNode> getLabelValues(Long folderId,Long groupByNodeId, List<Long> nodeIds, Long sortByNodeId);
+
+      List<JsonNode> getGroupedValues(Long nodeId, List<Long> filterNodeIds, Long sortByNodeId);
+
+
 }
--- a/src/main/java/io/hyperfoil/tools/h5m/rest/FolderResource.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/rest/FolderResource.java
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.databind.JsonNode;
 import io.hyperfoil.tools.h5m.api.Folder;
 import io.hyperfoil.tools.h5m.api.FolderSummary;
 import io.hyperfoil.tools.h5m.api.svc.FolderServiceInterface;
+import io.hyperfoil.tools.h5m.api.svc.ValueServiceInterface;
 import io.hyperfoil.tools.yaup.json.Json;
 import io.quarkus.security.Authenticated;
 import jakarta.annotation.security.PermitAll;
@@ -26,6 +27,9 @@ public class FolderResource {
 
     @Inject
     FolderServiceInterface folderService;
+
+    @Inject
+    ValueServiceInterface valueService;
 
     @GET
     @PermitAll
@@ -103,4 +107,17 @@ public class FolderResource {
     public Json structure(@PathParam("name") String name) {
         return folderService.structure(name);
     }
+
+    @GET
+    @Path("{id}/labelValues")
+    @PermitAll
+    @Operation(description = "Get metrics labels Values")
+    public List<JsonNode>getLabelValues(
+                    @PathParam("id") Long folderId,
+                    @QueryParam("groupById") Long groupById,
+                    @QueryParam("nodeIds") List<Long> nodeIds,
+                    @QueryParam("sortById") Long sortById)
+            {
+                return valueService.getLabelValues(folderId, groupById,nodeIds,sortById);
+            }
 }
--- a/src/main/java/io/hyperfoil/tools/h5m/svc/ValueService.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/svc/ValueService.java
@@ -3,6 +3,7 @@ package io.hyperfoil.tools.h5m.svc;
 import com.fasterxml.jackson.databind.JsonNode;
 import io.hyperfoil.tools.h5m.api.Value;
 import io.hyperfoil.tools.h5m.api.svc.ValueServiceInterface;
+import io.hyperfoil.tools.h5m.entity.FolderEntity;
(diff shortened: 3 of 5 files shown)


src/main/java/io/hyperfoil/tools/h5m/rest/FolderResource.java
```diff
@@ -103,4 +107,17 @@ public void recalculate(@PathParam("name") String name) {
...
+    @GET
+    @Path("{name}/labelValues")
+    @PermitAll
+    @Operation(description = "Get metrics labels Values")
+    public List<JsonNode>getLabelValues(
+                    @PathParam("name") String name,
+                    @QueryParam("groupbyId") Long groupbyId,
+                    @QueryParam("nodeId") List<Long> nodeId,
+                    @QueryParam("sortById") Long sortById)
+            {
```
stalep, line 120: Parameter naming suggestions:
- `nodeId` → `nodeIds` (it's a `List<Long>`, plural is clearer)
- `groupbyId` → `groupById` (camelCase to match `sortById`)

src/main/java/io/hyperfoil/tools/h5m/svc/ValueService.java
```diff
@@ -433,55 +436,79 @@ with recursive ancestor(vid) as (
...
-                    """.replace("NODE_FILTER", nodeFilter);
-                default -> "";
-            }, JsonNode.class
-        ).setParameter("nodeId", nodeId);
-        if (filterNodeIds != null) {
-            query.setParameter("nodeIds", filterNodeIds);
-        }
-        return query.getResultList();
+        return getGroupedValues(nodeId,filterNodeIds,null);
     }
```
stalep, line 441: The sort CTE uses `min(data::text)::jsonb` which sorts by text representation. This means numeric values sort lexicographically (`"100" < "20" < "200"`). For the test data with 3-digit numbers this works, but for mixed-length numbers it would produce wrong ordering. Worth documenting this limitation or considering a numeric cast when the data is numeric.

src/main/java/io/hyperfoil/tools/h5m/svc/ValueService.java
```diff
@@ -587,6 +614,26 @@ public List<Value> getNodeValues(Long nodeId){
...
+    @Transactional
+    public List<JsonNode> getLabelValues(String folderName, Long groupByNodeId, List<Long> nodeIds, Long sortByNodeId) {
+        FolderEntity folder = FolderEntity.find("name",folderName).firstResult();
+        if(folder == null){
+            throw new  NotFoundException("Folder not found: " + folderName);
+        }
+        Long rootNodeId=folder.group.root.id;
+        List<Long> filterNodeIds = new ArrayList<>();
+        if(nodeIds != null ){filterNodeIds.addAll(nodeIds);}
+        if(groupByNodeId != null){
```
stalep, line 627: `getLabelValues` always adds `groupByNodeId` and `sortByNodeId` to the filter list. This means if a user passes `nodeId=5&sortById=10`, the sort node appears as a column in the output even if the user only wanted it for ordering. Is this intentional? If so, worth documenting in the Javadoc.

src/main/java/io/hyperfoil/tools/h5m/api/svc/ValueServiceInterface.java
```diff
@@ -48,4 +48,17 @@ public interface ValueServiceInterface {
      */
     List<Value> getNodeValues(Long nodeId);
 
+    /**
+         * Returns one row per upload for a folder, with each row containing the values of the requested nodes.
+         * groupByNodeId and sortByNodeId are always included in every row regardless of nodeIds.
+         * @param folderName   Folder to query.
+         * @param nodeIds      Node IDs to include as columns (null = all nodes).
+         * @param groupByNodeId Node whose value identifies the series (e.g. config fingerprint).
```
stalep, line 56: Minor: the Javadoc indentation uses extra spaces compared to the rest of the file. Also could mention that an empty `nodeIds` list is treated the same as `null` (returns all nodes) — this is a behavior change from the original which would have generated an empty IN clause.

stalep (commented):
Good start on the labelValues API. The refactoring of getGroupedValues into a 3-param method with sort support is clean, and the test coverage is solid. A few suggestions below.

src/main/java/io/hyperfoil/tools/h5m/svc/ValueService.java
Arjun-Parmani: So Sort by id basically acts a domain node or time series along the X-axis. This helps to understand the graph in a better way. for example - Without build_id appearing as a column, Grafana would receive
rows in the right order but have no value to put on the X-axis.

src/main/java/io/hyperfoil/tools/h5m/api/svc/ValueServiceInterface.java
Arjun-Parmani: Rectified the indentation and added a mention - empty list of nodes = all nodes

src/main/java/io/hyperfoil/tools/h5m/rest/FolderResource.java
Arjun-Parmani: Rectified the naming conventions

src/main/java/io/hyperfoil/tools/h5m/svc/ValueService.java
Arjun-Parmani: Casted the data in numeric by checking if the data received is a number

src/main/java/io/hyperfoil/tools/h5m/rest/FolderResource.java
```diff
@@ -103,4 +107,17 @@ public void recalculate(@PathParam("name") String name) {
     public Json structure(@PathParam("name") String name) {
         return folderService.structure(name);
     }
+
+    @GET
+    @Path("{name}/labelValues")
```
barreiro: Should this use `folderId` instead of name ??

Arjun-Parmani: After the discussion, using folder Id and fetching folder based on the ID and not name. This will help to identify the folder uniquely even if the folder name is changed in future

PR #127, opened 2026-06-14
Ephemeral value data with three-state nullable Boolean (issue #123)

stalep:
Three-state ephemeral flag on NodeEntity:
- null  = auto (system can mark via markAutoEphemeral)
- true  = user explicitly wants data discarded
- false = user explicitly wants data kept

NodeEntity:
- ephemeral: Boolean (nullable)

ValueService:
- nullifyEphemeralData(rootValueId): nullifies data for nodes with ephemeral=true only. Root and detection nodes excluded as safety net. Called after upload and recovery completion.
- nullifyNodeData(nodeId): immediately nullifies all data for a node. Called when user sets ephemeral=true via REST.
- markAutoEphemeral(groupId): bulk-marks intermediate nodes (ephemeral IS NULL + has non-detection children) as ephemeral=true. Skips root, detection, and user-locked (false) nodes.

FolderService:
- upload() and recoverIncompleteUploads() both call nullifyEphemeralData in their whenComplete callbacks.

NodeResource:
- PUT /api/node/{id}/ephemeral?mode=true|false|auto
  - true: sets ephemeral=true + immediately nullifies existing data
  - false: sets ephemeral=false (user override)
  - auto: sets ephemeral=null
  - Detection and root nodes cannot be set to ephemeral

Node API record:
- ephemeral: Boolean (nullable)

Benchmark (10 rhivos runs):
- Upload time: 158s (no regression vs 160s on main)
- DB size: 100 MB (83% reduction from 586 MB on main)

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/entity/NodeEntity.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/entity/NodeEntity.java
@@ -1,6 +1,7 @@
 package io.hyperfoil.tools.h5m.entity;
 
 import com.fasterxml.jackson.annotation.*;
+import io.hyperfoil.tools.h5m.api.EphemeralMode;
 import io.hyperfoil.tools.h5m.api.NodeType;
 import io.hyperfoil.tools.h5m.entity.node.RootNode;
 import io.hyperfoil.tools.h5m.queue.KahnDagSort;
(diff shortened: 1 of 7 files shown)


src/main/java/io/hyperfoil/tools/h5m/entity/NodeEntity.java
```diff
@@ -29,6 +29,18 @@ public static enum ScalarVariableMethod { First, All}
...
+     * Controls whether value data for this node is nulled out after upload
+     * processing completes to save storage.
+     *
+     * null  = auto: ephemeral if node has non-detection children (intermediate node)
+     * true  = user explicitly wants data discarded
+     * false = user explicitly wants data kept
+     *
+     * The value rows and edges are always preserved for ancestry queries.
+     */
+    public Boolean ephemeral;
```
diegolovison: Maybe use an enum?

willr3: This is probably a safer idea than using `null` to treat a binary as a trinary data type :)

src/main/java/io/hyperfoil/tools/h5m/svc/ValueService.java
```diff
@@ -661,6 +661,86 @@ public String getHash(ValueEntity value) throws NoSuchAlgorithmException {
...
+    }
+
+    /**
+     * Nulls out all existing value data for a specific node.
+     * Called when a user explicitly sets ephemeral = true on a node.
+     *
+     * @return the number of values whose data was nulled
+     */
+    @Transactional
+    public int nullifyNodeData(long nodeId) {
```
willr3, line 737: I think this would potentially break any calculations in flight for that node. If one user is updating the node while an upload is in progress then I suspect this method call would break that upload.

stalep: Thanks, nice catch. That should be fixed now by guarding nullifyNodeData against concurrent uploads.

diegolovison (approved):
Looks good about the enum

PR #138, opened 2026-06-17
changed operationStrategy (match Horrreum), disable quinoa for tests

stalep:


barreiro:
I was thinking of renaming some methods instead. 

for example, `NodeResource#create` should have been `NodeResource#createNode`

but it's personal preference

stalep:
@barreiro updated now, can you have another look?

barreiro:
looks good. 

~~since we're on it, can you also add `org.eclipse.microprofile.openapi.annotations.tags.Tag` annotation to the resource classes ?? that way, the generated client methods will be grouped in distinct classes too.~~(`@Tag` annotations ar already present and the generated javacript already has a class for each rest resource)

barreiro (approved):

PR #140, opened 2026-06-17
Avoid pollutes user home directory

diegolovison:
Fixes https://github.com/Hyperfoil/h5m/issues/134

We decided to use the current working directory instead of `~/.local/share`

```
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  14.347 s
[INFO] Finished at: 2026-06-17T10:26:40-03:00
[INFO] ------------------------------------------------------------------------
```
```
dlovison:h5m$ alias h5m="java -jar target/cli/h5m.jar"
dlovison:h5m$ h5m add folder test
dlovison:h5m$ pwd
/home/dlovison/github/Hyperfoil/h5m
dlovison:h5m$ ls h5m*
h5m.db  h5m.db-shm  h5m.db-wal  h5m.iml
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,7 @@ dependency-reduced-pom.xml
 perf_test/states.yaml
 perf_test/qDup.jar
 jmh-result.*
+
+/h5m.db
+/h5m.db-shm
+/h5m.db-wal
--- a/src/main/java/io/hyperfoil/tools/h5m/provided/DatasourceConfiguration.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/provided/DatasourceConfiguration.java
@@ -51,19 +51,13 @@ public class DatasourceConfiguration {
     public DatasourceConfiguration() {
     }
 
-    public static String getPath(){
+    public static String getPath() {
         String rtrn = "";
-        if(System.getenv("H5M_PATH") !=null ){
+        if (System.getenv("H5M_PATH") != null) {
             rtrn = System.getenv("H5M_PATH");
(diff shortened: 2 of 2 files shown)


stalep:
LGTM, only thing I just came to think of is that we do not verify that `$H5M_PATH` is an actual directory.

stalep:
ah, and also add *.db to .gitignore?

diegolovison:
Done

stalep:
Still no verification that $H5M_PATH is a directory. Do you not think it's needed?

diegolovison:
I don't think that is needed.

willr3:
The stated goal of this PR is to avoid polluting the home directory but the code does more than change the path.  This change would turn `h5m` into a command that is path dependent by default. I for one do not want to have to be mindful of where I run `h5m` to be able to see the work I previously performed with the tool. 
I understand users could still use `H5M_PATH` to restore `h5m` to a system wide command like `systctl` and `iptables` but why are we introducing the split operating mode if the stated goal is to change the default value for `H5M_PATH`?

You mention not wanting to "pollute" the home folder. The home folder is the idiomatic location for user files. `ls -al ~` and you will likely see several "dot" files that configure how commands behave on your system. Would you rather `h5m` creates the database as a hidden dot file? Wanting h5m to use a different location is already supported by adding `export H5M_PATH=/tmp` in your `~/.bashrc` (or whatever shell you are using). No additional code nor split mode introduced into `h5m`.

PR #141, opened 2026-06-17
Streamline local development workflow

diegolovison:
This PR introduces a few improvements to make local development significantly easier and faster, specifically when iterating on the CLI alongside Quarkus Dev mode. It automates the build/run cycle.

cli and quarkus:dev work together


willr3:
This project uses `mvn` to build the artifacts so I do not think we should add configuration files for `make`.

PR #143, opened 2026-06-17
Inline AUTO ephemeral logic and nullify on graph change (issue #123)

stalep:
- nullifyEphemeralData: inline AUTO resolution at nullification time. AUTO nodes with non-detection children are now resolved on the fly without needing a separate markAutoEphemeral() call.
- NodeService: nullify source node data when a child node is created, making previously-leaf nodes ephemeral immediately on graph change.
- ViewService: automatically mark nodes as KEEP when added to a view via createView() or updateView().
- Tests: update CLI tests to assert on leaf values only (intermediate data is nullified by design). Disable 2 tests that require CLI ephemeral command support (deferred to aesh CLI port).

### DIFF
--- a/src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
+++ b/src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
@@ -349,9 +349,10 @@ public class H5mTest {
 
         LaunchResult result = results.getLast();
         assertTrue(result.getOutput().contains("Count: 3"));
-        assertTrue(result.getOutput().contains(" {\"bar\":{\"biz\":\"buz\"}} "),"result should contain .foo:" +result.getOutput());
-        assertTrue(result.getOutput().contains(" {\"biz\":\"buz\"} "),"result should contain .bar:" +result.getOutput());
-        assertTrue(result.getOutput().contains(" buz "),"result should contain .bar:" +result.getOutput());
+        // Intermediate nodes (foo, bar) have ephemeral=AUTO and their data is nullified
+        assertTrue(result.getOutput().contains(" null "),"intermediate values should be null:" +result.getOutput());
+        // Only the leaf node (biz) retains its data
+        assertTrue(result.getOutput().contains(" buz "),"result should contain leaf value buz:" +result.getOutput());
     }
     @Test
     public void upload_folder_list_values(QuarkusMainLauncher launcher) throws IOException {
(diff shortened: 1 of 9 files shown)


src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
```diff
@@ -107,6 +109,7 @@ public Long create(String name, Long groupId, NodeType type, String operation){
         }
         NodeEntity merged = em.merge(node);
         em.flush();
+        nullifyEphemeralSources(merged);
```
willr3: Making this a synchronous call on any method in the REST API will result in a degrading performance over time. If nullifyEphemeralSources is called with 1000 uploads then creating a new node will wait for all 1000 to be updated.

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
```diff
@@ -349,9 +349,9 @@ public void upload_list_values(QuarkusMainLauncher launcher) throws IOException
 
         LaunchResult result = results.getLast();
         assertTrue(result.getOutput().contains("Count: 3"));
-        assertTrue(result.getOutput().contains(" {\"bar\":{\"biz\":\"buz\"}} "),"result should contain .foo:" +result.getOutput());
-        assertTrue(result.getOutput().contains(" {\"biz\":\"buz\"} "),"result should contain .bar:" +result.getOutput());
-        assertTrue(result.getOutput().contains(" buz "),"result should contain .bar:" +result.getOutput());
+        // Intermediate nodes (foo, bar) have ephemeral=AUTO and their data is nullified
+        // Only the leaf node (biz) retains its data
```
willr3, line 354: I believe this should be changed to an assert that the ephemeral values are missing and not simply removed

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
```diff
@@ -398,12 +398,10 @@ public void upload_folder_list_values(QuarkusMainLauncher launcher) throws IOExc
 
         LaunchResult result = results.getLast();
         assertTrue(result.getOutput().contains("Count: 6"));
-        assertTrue(result.getOutput().contains(" {\"bar\":{\"biz\":\"buz\"}} "),"result should contain .foo:" +result.getOutput());
-        assertTrue(result.getOutput().contains(" {\"bar\":{\"biz\":\"bur\"}} "),"result should contain .foo:" +result.getOutput());
-        assertTrue(result.getOutput().contains(" {\"biz\":\"buz\"} "),"result should contain .bar:" +result.getOutput());
-        assertTrue(result.getOutput().contains(" {\"biz\":\"bur\"} "),"result should contain .bar:" +result.getOutput());
-        assertTrue(result.getOutput().contains(" buz "),"result should contain .bar:" +result.getOutput());
-        assertTrue(result.getOutput().contains(" bur "),"result should contain .bar:" +result.getOutput());
+        // Intermediate nodes (foo, bar) have ephemeral=AUTO and their data is nullified
```
willr3, line 402: This too should have been changed to assert that ephemeral values were removed instead of simply removing the asserts

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
```diff
@@ -439,9 +437,9 @@ public void upload_jsonata_list_values(QuarkusMainLauncher launcher) throws IOEx
 
         LaunchResult result = results.getLast();
         assertTrue(result.getOutput().contains("Count: 3"),"expect 3 values\n"+result.getOutput());
-        assertTrue(result.getOutput().contains(" {\"bar\":{\"biz\":\"buz\"}} "),"result should contain .foo:" +result.getOutput());
-        assertTrue(result.getOutput().contains(" {\"biz\":\"buz\"} "),"result should contain .bar:" +result.getOutput());
-        assertTrue(result.getOutput().contains(" buz "),"result should contain .bar:" +result.getOutput());
+        // Intermediate nodes (foo, bar) have ephemeral=AUTO and their data is nullified
```
willr3, line 442: removing the asserts decreased the accuracy of the testing instead of updating the tests to remain accurate for the new outcome

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
```diff
@@ -477,9 +475,9 @@ public void upload_sqlpath_list_values(QuarkusMainLauncher launcher) throws IOEx
...
```
willr3, line 481: We should be testing that intermediate nodes are removed

src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
stalep: I ended up with that since we talked about a way where we only would remove ephemeral data in the next upload and the consensus was that we should look at removing it when the user selected it to be removed. I wasn't too fond of nullifying in another tx or thread so I've now gone back to the logic where we only remove during the next upload.

PR #144, opened 2026-06-18
Implement e-divisive (Hunter) change detection using jhunter (issue #…

stalep:
…114)

- Add jhunter-core 0.1.0 dependency
- Complete EDivisive entity: implements DetectionNode, config parsing, setNodes() with sources [fingerprint, groupBy, range, domain]
- EDivisiveConfig record: windowLen, maxPvalue, minMagnitude, maxSeriesLength, fingerprintFilter
- calculateEDivisiveValues() in NodeService: collects full range value series per fingerprint, limits to maxSeriesLength, deletes existing change points, calls jhunter Analysis.computeChangePoints()
- Output includes: index, pvalue, meanBefore/After, stdBefore/After, magnitude, hazardLevel (log ratio for triage sorting)
- AddEDivisive CLI command following AddRelativeDifference pattern
- Workaround for yaup Json.getDouble() BigDecimal bug (yaup#41)
- 7 tests: step function, flat series, multiple change points, insufficient data, recomputation, REST createConfigured, hazard level

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/cli/AddEDivisive.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/cli/AddEDivisive.java
@@ -1,0 +1,150 @@
+package io.hyperfoil.tools.h5m.cli;
+
+import io.hyperfoil.tools.h5m.api.EDivisiveConfig;
+import io.hyperfoil.tools.h5m.api.Node;
+import io.hyperfoil.tools.h5m.api.NodeGroup;
+import io.hyperfoil.tools.h5m.api.NodeType;
+import io.hyperfoil.tools.h5m.api.svc.NodeGroupServiceInterface;
+import io.hyperfoil.tools.h5m.api.svc.NodeServiceInterface;
+import io.hyperfoil.tools.h5m.entity.node.EDivisive;
+import jakarta.inject.Inject;
+import picocli.CommandLine;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.stream.Collectors;
+
+@CommandLine.Command(name = "edivisive", separator = " ", description = "add an e-divisive (Hunter) change detection node", mixinStandardHelpOptions = true)
+public class AddEDivisive implements Callable<Integer> {
+
+    @CommandLine.Option(names = {"to"}, description = "target group / test")
+    String groupName;
+
+    @CommandLine.Option(names = {"range"}, arity = "1", description = "node that produces the value to inspect")
+    String rangeName;
+
+    @CommandLine.Option(names = {"domain"}, arity = "1", required = true, description = "node used to sort the range values (required for e-divisive)")
+    String domainName;
+
+    @CommandLine.Option(names = {"windowLen"}, arity = "0..1", description = "sliding window size for the split phase (min 3)",
+            defaultValue = "" + EDivisive.DEFAULT_WINDOW_LEN)
+    int windowLen;
+
+    @CommandLine.Option(names = {"maxPvalue"}, arity = "0..1", description = "significance threshold for change points",
+            defaultValue = "" + EDivisive.DEFAULT_MAX_PVALUE)
+    double maxPvalue;
+
+    @CommandLine.Option(names = {"minMagnitude"}, arity = "0..1", description = "minimum relative change magnitude to report (e.g., 0.1 = 10%)",
+            defaultValue = "" + EDivisive.DEFAULT_MIN_MAGNITUDE)
+    double minMagnitude;
+
+    @CommandLine.Option(names = {"maxSeriesLength"}, arity = "0..1", description = "maximum number of recent data points to analyze",
+            defaultValue = "" + EDivisive.DEFAULT_MAX_SERIES_LENGTH)
+    int maxSeriesLength;
+
+    @CommandLine.Option(names = {"fingerprint"}, description = "node names to use as fingerprint")
+    List<String> fingerprints;
+
+    @CommandLine.Option(names = {"--fingerprint-filter", "-ff"}, arity = "0..1", description = "jq filter expression for fingerprints")
+    String fingerprintFilter;
+
+    @CommandLine.Option(names = {"by"}, description = "grouping node", arity = "0..1")
+    public String groupBy;
+
+    @CommandLine.Parameters(index = "0", arity = "1", description = "node name")
+    String name;
+
+    @Inject
+    NodeGroupServiceInterface nodeGroupService;
+
+    @Inject
+    NodeServiceInterface nodeService;
+
+    @Override
+    public Integer call() throws Exception {
+        if (name == null || name.isEmpty()) {
+            System.err.println("missing node name");
+            return 1;
+        }
+        if (groupName == null || groupName.isEmpty()) {
+            System.err.println("missing group name");
+            return 1;
+        }
+        NodeGroup foundGroup = nodeGroupService.byName(groupName);
+        if (foundGroup == null) {
+            System.err.println("node group with name " + groupName + " does not exist");
+            return 1;
+        }
+
+        List<Node> foundNodes = nodeService.findNodeByFqdn(name, foundGroup.id());
+        if (!foundNodes.isEmpty()) {
+            System.err.println(groupName + " already has " + name + " node(s)\n  " + foundNodes.stream().map(Node::fqdn).collect(Collectors.joining("\n  ")));
+        }
+
+        if (rangeName == null || rangeName.isEmpty()) {
+            System.err.println("Missing range");
+            return 1;
+        }
+        foundNodes = nodeService.findNodeByFqdn(rangeName, foundGroup.id());
+        if (foundNodes.isEmpty()) {
+            System.err.println("could not find matching range node by name " + rangeName);
+            return 1;
+        } else if (foundNodes.size() > 1) {
+            System.err.println("found more than one matching range node by name " + rangeName + "\n  " + foundNodes.stream().map(Node::fqdn).collect(Collectors.joining("\n  ")));
+            return 1;
+        }
+        Node rangeNode = foundNodes.getFirst();
+
+        // domainName is required=true, so always present
+        foundNodes = nodeService.findNodeByFqdn(domainName, foundGroup.id());
+        if (foundNodes.isEmpty()) {
+            System.err.println("could not find matching domain node by name " + domainName);
+            return 1;
+        } else if (foundNodes.size() > 1) {
+            System.err.println("found more than one matching domain node by name " + domainName + "\n  " + foundNodes.stream().map(Node::fqdn).collect(Collectors.joining("\n  ")));
+            return 1;
+        }
+        Node domainNode = foundNodes.getFirst();
+
+        Node groupByNode = null;
+        if (groupBy != null && !groupBy.isEmpty()) {
+            foundNodes = nodeService.findNodeByFqdn(groupBy, foundGroup.id());
+            if (foundNodes.isEmpty()) {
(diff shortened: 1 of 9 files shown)


src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
```diff
@@ -676,6 +689,122 @@ public List<ValueEntity> calculateFixedThresholdValues(FixedThreshold ft, ValueE
...
+                }
+
+                // Collect all range values for this fingerprint, ordered by domain (or created_at)
+                List<ValueEntity> rangeValues;
+                if (ed.getDomainNode() != null) {
+                    rangeValues = valueService.findMatchingFingerprint(
+                            ed.getRangeNode(), groupBy, fingerprintValue, ed.getDomainNode());
+                } else {
+                    rangeValues = valueService.findMatchingFingerprint(
+                            ed.getRangeNode(), groupBy, fingerprintValue, (NodeEntity) null);
```
willr3: This is going to fetch all range values ordered by `created_at` but without any starting value.

src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
```diff
@@ -676,6 +689,122 @@ public List<ValueEntity> calculateFixedThresholdValues(FixedThreshold ft, ValueE
...
+                }
+
+                // Limit series to most recent maxSeriesLength values
+                if (maxSeriesLength > 0 && rangeValues.size() > maxSeriesLength) {
+                    rangeValues = rangeValues.subList(rangeValues.size() - maxSeriesLength, rangeValues.size());
```
willr3: This will potentially select values that do not include the upload if the user did not specify a `domainNode` for `ed`.

src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
```diff
@@ -676,6 +689,122 @@ public List<ValueEntity> calculateFixedThresholdValues(FixedThreshold ft, ValueE
...
+                }
+
+                // Extract double[] from range values
+                double[] series = new double[rangeValues.size()];
+                int validCount = 0;
+                for (int i = 0; i < rangeValues.size(); i++) {
+                    ValueEntity rv = rangeValues.get(i);
+                    if (rv.data instanceof NumericNode numericNode) {
+                        series[validCount++] = numericNode.asDouble();
+                    } else if (rv.data != null && rv.data.asText().matches("-?[0-9]+\\.?[0-9]*")) {
```
willr3: This lacks a size bound check on the regex. 100 characters will not fit into Double.

src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
```diff
@@ -676,6 +689,122 @@ public List<ValueEntity> calculateFixedThresholdValues(FixedThreshold ft, ValueE
...
+                        series[validCount++] = Double.parseDouble(rv.data.asText());
```
willr3: Should this catch `NumberFormatException` so that the exception does not abort the calculation?

src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
```diff
@@ -676,6 +689,122 @@ public List<ValueEntity> calculateFixedThresholdValues(FixedThreshold ft, ValueE
...
+                if (validCount < ed.getWindowLen()) {
+                    continue;
+                }
+
+                // Trim array if some values were non-numeric
+                if (validCount < series.length) {
+                    series = Arrays.copyOf(series, validCount);
+                }
+
+                // Delete existing e-divisive values for this fingerprint scope
```
willr3: If this is going to delete all past e-divisive change detections then I think we need to make this a `cumulative` Work and not merge the PR I just introduced that removed `cumulative` from Work :)

src/test/java/io/hyperfoil/tools/h5m/svc/EDivisiveTest.java
```diff
@@ -0,0 +1,318 @@
...
+        // Series with a clear step: 10 values at mean=100, then 10 values at mean=200
+        double[] series = new double[20];
+        for (int i = 0; i < 10; i++) series[i] = 100.0 + (i % 3); // slight noise
+        for (int i = 10; i < 20; i++) series[i] = 200.0 + (i % 3);
+
+        long edId = setupAndUpload("step-test", series, 5, 0.05);
+
+        tm.begin();
+        List<ValueEntity> edValues = ValueEntity.find("node.id", edId).list();
+        assertFalse(edValues.isEmpty(), "Should detect at least one change point");
```
willr3: This should instead be asserting for the expected number of changes, if this suddenly returns 50 changes then the test should fail.

src/test/java/io/hyperfoil/tools/h5m/svc/EDivisiveTest.java
```diff
@@ -0,0 +1,318 @@
...
+        int cpIndex = cp.data.get("index").asInt();
+        assertTrue(cpIndex >= 8 && cpIndex <= 12,
+                "Change point should be near index 10, was " + cpIndex);
+
+        // meanBefore should be around 100, meanAfter around 200
+        double meanBefore = cp.data.get("meanBefore").asDouble();
+        double meanAfter = cp.data.get("meanAfter").asDouble();
+        assertTrue(meanBefore < 150, "meanBefore should be near 100, was " + meanBefore);
+        assertTrue(meanAfter > 150, "meanAfter should be near 200, was " + meanAfter);
+
```
willr3, line 153: I think we should have asserts that verify the sources of the `cp` are the expected sources. Otherwise this is verifying the changepoint is occurring at the expected location.

src/test/java/io/hyperfoil/tools/h5m/svc/EDivisiveTest.java
```diff
@@ -0,0 +1,318 @@
...
+        assertTrue(cp.data.has("index"), "Should have index field");
+        assertTrue(cp.data.has("meanBefore"), "Should have meanBefore field");
+        assertTrue(cp.data.has("meanAfter"), "Should have meanAfter field");
+        assertTrue(cp.data.has("pvalue"), "Should have pvalue field");
+        assertTrue(cp.data.has("magnitude"), "Should have magnitude field");
+        assertTrue(cp.data.has("fingerprint"), "Should have fingerprint field");
+
+        // The change point should be near index 10
+        int cpIndex = cp.data.get("index").asInt();
+        assertTrue(cpIndex >= 8 && cpIndex <= 12,
```
willr3: Why are you using a range for the changepoint? Is the algorithm non-deterministic for the inputs?

src/test/java/io/hyperfoil/tools/h5m/svc/EDivisiveTest.java
```diff
@@ -0,0 +1,318 @@
...
+        double[] series = new double[30];
+        for (int i = 0; i < 10; i++) series[i] = 100.0;
+        for (int i = 10; i < 20; i++) series[i] = 200.0;
+        for (int i = 20; i < 30; i++) series[i] = 50.0;
+
+        long edId = setupAndUpload("multi-cp-test", series, 5, 0.05);
+
+        tm.begin();
+        List<ValueEntity> edValues = ValueEntity.find("node.id", edId).list();
+        assertTrue(edValues.size() >= 2,
```
willr3: Why do you need to use a range rather than asserting for the expected number of changes?

willr3 (commented):
You created a single `EDivisiveTest` to unit test multiple different classes where we previously located united tests in a `XTest` for each class X. Do you think we should keep the `XTest` per class or move unit tests to a unified class for each Node?

src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
stalep: Replaced the two-branch query + Java truncation with a single findMatchingFingerprint call that passes maxSeriesLength as the SQL LIMIT.

src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
stalep: Added size check

src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
stalep: Wrapped in try-catch.

src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
stalep: I added a todo in the code. I think we should discuss this further

src/test/java/io/hyperfoil/tools/h5m/svc/EDivisiveTest.java
stalep: fixed

src/test/java/io/hyperfoil/tools/h5m/svc/EDivisiveTest.java
stalep: fixed

src/test/java/io/hyperfoil/tools/h5m/svc/EDivisiveTest.java
stalep: changed to exact count

stalep:
> You created a single `EDivisiveTest` to unit test multiple different classes where we previously located united tests in a `XTest` for each class X. Do you think we should keep the `XTest` per class or move unit tests to a unified class for each Node?

I think a dedicated test class per detection algorithm makes sense — each algorithm has distinct behavior to verify (e.g., e-divisive's delete-and-recompute, multiple change points, series length handling).

src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
stalep: I think we need to look at recalculation and how that should be done since it ties a lot with this.

src/main/java/io/hyperfoil/tools/h5m/cli/AddEDivisive.java
```diff
@@ -0,0 +1,154 @@
...
+                } else if (foundNodes.size() > 1) {
+                    System.err.println("found more than one matching fingerprint node by name " + fingerprintName + "\n  " + foundNodes.stream().map(Node::fqdn).collect(Collectors.joining("\n  ")));
+                    return 1;
+                }
+                fingerprintNodes.add(foundNodes.getFirst().id());
+            }
+        }
+
+        Long fingerprintId = nodeService.createConfigured("_fp-" + name, foundGroup.id(), NodeType.FINGERPRINT, fingerprintNodes, null);
+        List<Long> sources = domainNode == null
```
Arjun-Parmani: In the above code the domain node is declared as `required = true`. Does it makes this code redundant?

src/main/java/io/hyperfoil/tools/h5m/cli/AddEDivisive.java
```diff
@@ -0,0 +1,154 @@
...
+            System.err.println("could not find matching range node by name " + rangeName);
+            return 1;
+        } else if (foundNodes.size() > 1) {
+            System.err.println("found more than one matching range node by name " + rangeName + "\n  " + foundNodes.stream().map(Node::fqdn).collect(Collectors.joining("\n  ")));
+            return 1;
+        }
+        Node rangeNode = foundNodes.getFirst();
+
+        Node domainNode = null;
+        if (domainName != null && !domainName.isEmpty()) {
```
Arjun-Parmani: This block will always be true since domainName is declared as `required = true` above.  Do you think we still need this check?

PR #146, opened 2026-06-18
Add fallback for raw JSON parsing

diegolovison:


willr3:
I can see where typing json into the terminal is convenient for experimenting with `h5m` but I do not think this is a feature users would want when using `h5m` with more realistic performance data. The use case feels limited to experimentation and therefore I don't think we should add it. I also do not agree with  adding this feature so that we could include it in the README. Using json as a string in the README examples would not be how users upload data when working with performance testing data so I do not think we should teach them one use case to then have them use a different use case when using `h5m`.

stalep:
I'm closing this for now, but I think we should add this feature when we've ported the cli to aesh.

PR #148, opened 2026-06-18
remove cumulative from Work

willr3:
Remove cumulative from Work so that work does not get sorted to the back of the queue


stalep:
closing as discussed

PR #149, opened 2026-06-19
Fix recalculate and add selective recalculate with auto-trigger (issu…

stalep:
…e #142)

- Use top-level nodes + cascade (like upload), not all-nodes Cartesian
- Return CompletableFuture so callers can wait for completion
- Run nullifyEphemeralData in whenComplete callback
- Add dispatch flag to Work for notification suppression
- Wire dispatch through WorkService.execute() cascade to ChangeDetectedEvent
- FolderService.recalculateNode(name, nodeId): recomputes a specific node and its dependents via cascade. Only the target node is queued per root value. Dedup in WorkService.execute() skips unchanged values.
- NodeService.update(): auto-triggers recalculateNode when a node's operation changes
- REST endpoint: POST /api/folder/{name}/recalculate/{nodeId}
- FolderServiceInterface: updated with new method signatures


src/main/java/io/hyperfoil/tools/h5m/rest/FolderResource.java
```diff
@@ -96,6 +96,14 @@ public void recalculate(@PathParam("name") String name) {
         folderService.recalculate(name);
     }
 
+    @POST
+    @Path("{name}/recalculate/{nodeId}")
+    @Authenticated
+    @Operation(description = "Recalculate values for a specific node and its dependents")
+    public void recalculateNode(@PathParam("name") String name, @PathParam("nodeId") Long nodeId) {
+        folderService.recalculateNode(name, nodeId);
```
Arjun-Parmani: The folderService.recalculateNode() calls a method in the interface which has a return type of CompletableFuture<Void>. But the API has a void return type? Why is there a mismatch of the return type? Was this intentional?

src/main/java/io/hyperfoil/tools/h5m/svc/FolderService.java
```diff
@@ -287,27 +287,124 @@ public Json structure(String name) {
...
+            // Queue only top-level nodes for each root value (same as upload).
+            // Cascade in WorkService.execute() handles downstream nodes.
+            List<NodeEntity> topLevelNodes = folder.group.getTopLevelNodes();
+            List<Work> works = new ArrayList<>();
+            Set<Long> rootValueIds = new HashSet<>();
+            for (ValueEntity rootValue : rootValues) {
+                rootValueIds.add(rootValue.id);
+                for (NodeEntity node : topLevelNodes) {
+                    Work w = new Work(node, new ArrayList<>(node.sources), List.of(rootValue));
+                    w.dispatch = false; // suppress external notifications during recalculation
```
Arjun-Parmani: What happens when the user calls POST /api/folder/{name}/recalculate/{nodeId} while an upload for the same folder is still being processed? The work queue can end up with two Work items targeting the
 same node and the same root value, one from the upload with dispatch=true, one from the recalculate with dispatch=false. If the recalculate enters the queue first, it can drop the upload work will be dropped and if there is change detection it will not trigger any notification since dispatch will be false.

src/main/java/io/hyperfoil/tools/h5m/rest/FolderResource.java
stalep: Yes, it was, but when you mention it I think we should probably return a CompletionStage<Void> :+1:

src/main/java/io/hyperfoil/tools/h5m/svc/FolderService.java
stalep: Ah, good catch. I need to think about that.... Perhaps if we notice a dedup and one of the items has `dispatch=true` we should set all to true?

Arjun-Parmani: If we set all to true then there can be a scenario where recalculated node has arrived first with "dispatch=false". Updating it to "dispatch=true" will trigger notification if a change is detected for this node. This nullifies the implementation here.

stalep: Yeah, so now both recalculate and recalculateNode check `UploadProcessingEntity.count("folderName = ?1 and completed = false", name)` before creating Work items. Throws IllegalStateException if uploads are in progress.

src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
```diff
@@ -176,6 +177,13 @@ public long update(NodeEntity node){
                 }
             }
             em.merge(node);
+            em.flush();
+
+            // Auto-trigger selective recalculation when the node's operation changes
+            if (operationChanged && existing.group != null) {
+                String folderName = existing.group.name;
+                folderService.recalculateNode(folderName, node.id);
```
Arjun-Parmani: NodeService.update() calls recalculateNode which throws IllegalStateException If an upload is in progress. Since NodeService.update() is @Transactional, it can be possible if an exception has occured and the transaction is rolledback. Do you think in that case the node is updated?

stalep: The `update()` method itself doesn't call `recalculateNode()`, they are separate operations. But you raise a good point: if we ever chain them (e.g., in UpdateNode CLI or a REST endpoint), we should ensure the node update commits in its own transaction before triggering recalculation, so a rejection doesn't roll back the update.

src/main/java/io/hyperfoil/tools/h5m/rest/NodeResource.java
```diff
@@ -64,11 +74,46 @@ public Long createConfigured(
...
 
+    public record NodeUpdateRequest(String name, String operation) {}
+    public record NodeUpdateResponse(long nodeId, RecalculationStatus recalculation) {}
+
+    @PUT
+    @Path("{id}")
+    @Authenticated
+    @Operation(description = "Update a node's name and/or operation. If the operation changes, triggers selective recalculation.")
+    public NodeUpdateResponse update(@PathParam("id") Long id, NodeUpdateRequest request) {
+        NodeEntity existing = NodeEntity.findById(id);
```
Arjun-Parmani, line 85: when nodeServiceImpl.update(existing) calls NodeService.update, there is again a same check NodeEntity existing = NodeEntity.findById(node.id). Why are we fetching it twice?

src/main/java/io/hyperfoil/tools/h5m/entity/work/Work.java
```diff
@@ -25,6 +24,10 @@ public class Work implements Runnable, Comparable<Work>{
 
     public boolean cumulative = false;
 
+    /** Whether external notifications should be dispatched for detection results.
+     *  Set to false for recalculations and bulk imports. */
+    public boolean dispatch = true;
```
willr3, line 30: Why would a user want to disable notification on a recalculation? The de-duplication should already prevent re-notification for the same change detection (no new value will be persisted) but a new change detection that occurs from a change to a Node that triggered a recalculate is a valid change detection. How else do we plan for h5m to notify users that changing a node resulted in different changes being detected?

src/main/java/io/hyperfoil/tools/h5m/entity/ProcessingTrackerEntity.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+@Entity(name = "processing_tracker")
+public class ProcessingTrackerEntity extends PanacheEntity {
+
+    @Enumerated(EnumType.STRING)
+    @Column(nullable = false, updatable = false)
+    public ProcessingType type;
+
+    @Column(name = "folder_name", nullable = false, updatable = false)
+    public String folderName;
```
willr3: This should probably be a reference to the FolderEntity and not a detached reference to a mutable attribute on the Folder.

src/main/java/io/hyperfoil/tools/h5m/rest/FolderResource.java
```diff
@@ -66,38 +70,56 @@ public Map<String, Integer> getFolderUploadCount() {
...
     }
 
     @POST
     @Path("{name}/upload")
     @Authenticated
-    @Operation(description = "Upload JSON data to a folder")
-    public void upload(
+    @Blocking
+    @Operation(description = "Upload JSON data to a folder. Returns when all processing completes.")
+    public CompletionStage<Void> upload(
```
willr3, line 94: Is this the return type to a user who `curl -X POST /upload`? I would assume the return value should be the id of the uploaded value.

src/main/java/io/hyperfoil/tools/h5m/svc/WorkService.java
```diff
@@ -213,6 +213,13 @@ public CompletableFuture<Void> createTracked(List<Work> works, Set<Long> rootVal
         return CompletableFuture.allOf(futures.toArray(CompletableFuture[]::new));
     }
 
+    /**
+     * Returns the tracker for the given root value ID, if one exists.
+     */
+    public java.util.Optional<io.hyperfoil.tools.h5m.queue.UploadTracker> getTracker(long rootValueId) {
```
willr3: why is the fully qualified name for UploadTracker used here instead of `import`?

src/main/java/io/hyperfoil/tools/h5m/svc/FolderService.java
```diff
@@ -72,49 +78,163 @@ private static LocalDateTime toLocalDateTime(Object value) {
...
     }
 
+    private void recoverUpload(ProcessingTrackerEntity tracking) {
+        ValueEntity rootValue = ValueEntity.findById(tracking.referenceId);
+        if (rootValue == null) {
+            Log.warnf("Root value %d not found for incomplete upload, removing tracking record", tracking.referenceId);
+            tracking.delete();
+            return;
+        }
+        FolderEntity folder = findFolderByName(tracking.folderName);
```
willr3: I mentioned earlier but fetching by a mutable attribute is waiting for a concurrent modification to break it.

src/main/java/io/hyperfoil/tools/h5m/svc/FolderService.java
```diff
@@ -72,49 +78,163 @@ private static LocalDateTime toLocalDateTime(Object value) {
...
+        // A mid-process crash may have left the pipeline in a partially computed
+        // state — e.g., node A was recomputed but cascade to B, C didn't happen.
+        // With top-level + cascade, A's dedup sees the value is already correct
+        // and skips it without cascading. B and C remain stale.
+        //
+        // Why not use recalculateNode():
+        // Same dedup issue — if the tracked node's value already matches, cascade
+        // doesn't fire for its dependents.
+        //
+        // Solution: queue Work for EVERY node in the graph (same as upload recovery).
```
willr3: This solution will work but will result in the same wasted efforts that lead us to stop using this approach for upload. Would `findRecomputationStartNodes` be usable in this context too but instead of fetching the source ephemeral node we fetch all the ephemeral nodes in the source path?

src/main/java/io/hyperfoil/tools/h5m/svc/FolderService.java
```diff
@@ -287,27 +407,265 @@ public Json structure(String name) {
...
-                FolderEntity.class
-        ).setParameter("name", name).getSingleResult();
-        NodeEntity root = folder.group.root;
-        List<ValueEntity> rootValues = valueService.getValues(root);
-        rootValues.forEach(ValueEntity::getPath); // this fixes the LazyException, por que?
-        List<Work> newWorks = new ArrayList<>();
-        for(ValueEntity rootValue: rootValues){
-            for(NodeEntity source : List.copyOf(folder.group.sources)){
-                newWorks.add(new Work(source,new ArrayList<>(source.sources),List.of(rootValue)));
+    public RecalculationStatus recalculate(String name) {
```
willr3: I think we can drop this method from the API unless you can think of a user use case where they would want to trigger this behavior?

src/main/java/io/hyperfoil/tools/h5m/svc/FolderService.java
```diff
@@ -287,27 +407,265 @@ public Json structure(String name) {
...
+     * intermediate nodes before reaching the target.
+     *
+     * Use case: user changes a JQ expression or detection config for a node.
+     *
+     * @param name the folder name
+     * @param nodeId the node to recalculate
+     * @return future that completes when recalculation finishes
+     * @throws IllegalStateException if uploads are in progress for this folder
+     */
+    public RecalculationStatus recalculateNode(String name, long nodeId) {
```
willr3: If the user is providing the nodeId and nodes are not shared between folders then I think `name` is an unnecessary parameter. The folder can be fetched from the Node

src/main/java/io/hyperfoil/tools/h5m/api/RecalculationStatus.java
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Tracks the progress of a recalculation operation.
+ * Stored in-memory (not persisted) — lost on server restart.
+ * Serializable as JSON for REST responses (internal fields excluded via @JsonIgnore).
+ */
+public class RecalculationStatus {
+
+    public enum State { RUNNING, COMPLETED, FAILED }
+
+    private final String id;
```
willr3: what is the `id` that is a `string`? Is this just some uuid generated for the Recalculation?

willr3 (changes requested):

src/main/java/io/hyperfoil/tools/h5m/entity/work/Work.java
stalep: In Horreum today there is no notification on recalculation afaik as it can spam the group with notifications. - but we can change it h5m to behave differently... Actually it might be something the users should be able to configure in the future.

src/main/java/io/hyperfoil/tools/h5m/rest/FolderResource.java
stalep: I think this should be handled by #159

src/main/java/io/hyperfoil/tools/h5m/svc/FolderService.java
stalep: This can be fairly complicated as we do not know where in the recalculation we crashed. This can leave cascade gaps (dedup prevent cascade when values match). I think we can create an issue to optimize it later.

src/main/java/io/hyperfoil/tools/h5m/svc/FolderService.java
stalep: I agree, we can restrict it to nodeId

src/main/java/io/hyperfoil/tools/h5m/svc/FolderService.java
stalep: Yes, see above

src/main/java/io/hyperfoil/tools/h5m/rest/FolderResource.java
willr3: If we are deferring the change to that PR then I believe this PR should not change the API to an intermediary state that we plan to change yet again in another PR. Wouldn't that introduce an un-necessary merge conflict for whomever is working on #159?

src/main/java/io/hyperfoil/tools/h5m/api/RecalculationStatus.java
stalep: yes, uuid generated in RecalculationService. Used for Map lookup.

src/main/java/io/hyperfoil/tools/h5m/rest/NodeResource.java
stalep: Thanks, fixed!

src/main/java/io/hyperfoil/tools/h5m/entity/work/Work.java
stalep: created #160

src/main/java/io/hyperfoil/tools/h5m/rest/FolderResource.java
stalep: It's common that an issue depends on another?
 The` CompletionStage<Void>` return type is intentional as a foundation for #159 — that issue will build on top of this branch since we already have the infrastructure it needs:
ProcessingTrackerEntity with ProcessingType.UPLOAD already tracks upload lifecycle with crash recovery
RecalculationTracker/RecalculationStatus pattern (internal mutable tracker + pure record DTO snapshot) is exactly the what Arjun described in #159 for upload status tracking.

src/main/java/io/hyperfoil/tools/h5m/svc/FolderService.java
stalep: Created #161

PR #163, opened 2026-07-02
Upload REST end point in FolderResource now returns Upload Id immediately

Arjun-Parmani:
### Problem statement
     
  The /api/folder/{name}/upload endpoint was synchronous — it blocked until all node processing (JQ transformations, change detection, etc.) completed before returning, meaning:
  - No way to identify which upload triggered which processing
  - No way to correlate an upload to its ProcessingTrackerEntity record for crash recovery
  - Long-running uploads would hold the HTTP connection open                                                                     

  ### Approach

  Added UploadHandle.java — a single class that carries both the upload identity and the processing future (Similar pattern to that of Recalculation):

  ```
public class UploadHandle {
      public final long uploadId;       // serialized to client
      @JsonbTransient
      public final CompletableFuture<Void> future;  // internal use only
  }
```

  Refactored the core upload logic into a private executeUpload() method that returns UploadHandle. Two public methods delegate to it:
  - upload() (existing interface method) — extracts .future for java method callers like LoadLegacyRuns that need to await completion
  - uploadGetId() (new) — returns the full UploadHandle for the REST layer

  The REST endpoint now mirrors the recalculate pattern — returns UploadHandle directly:

  public UploadHandle upload(...) {
      if (data == null) throw new BadRequestException("Missing request body");
      return folderService.uploadGetId(name, path, data);
  }

  Also fixed a bug where entity.completed = true was set unconditionally in whenComplete, even on failure. Now only set on success (t == null).

  ### Outcome

  - Upload returns immediately example -  {"uploadId": 42}
  - uploadId is the ValueEntity ID of the uploaded root value — directly queryable in the DB
  - Existing callers (LoadLegacyRuns, CLI) are unaffected — upload() signature unchanged
  - Added tests in RestEndpointTest and FolderServiceTest covering: positive ID, uniqueness, tracker creation, ID-to-value match, empty body rejection

### To test via swagger

1. Enter into dev moe. - mvn quarkus:dev
2. route to swagger-ui page - http://localhost:8080/q/swagger-ui/
3. create a folder first - POST /api/folder/{test-uploadId}
4. upload with body using same folder name - POST /api/folder/test-uploadId/upload. - body - {"cpu": 92}
5. Response received - Response body{   "uploadId": 201 }

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/api/svc/FolderServiceInterface.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/api/svc/FolderServiceInterface.java
@@ -3,13 +3,13 @@ package io.hyperfoil.tools.h5m.api.svc;
 import io.hyperfoil.tools.jjq.value.JqValue;
 import io.hyperfoil.tools.h5m.api.Folder;
 import io.hyperfoil.tools.h5m.api.FolderSummary;
+import io.hyperfoil.tools.h5m.api.Upload;
 import io.hyperfoil.tools.h5m.svc.RecalculationTracker;
 
 import java.io.IOException;
 import java.nio.file.Path;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.CompletableFuture;
 
 /**
  * Service interface for managing Folders.
@@ -56,15 +56,16 @@ public interface FolderServiceInterface {
 
     /**
      * Uploads data to a specific path within a folder.
-     * Returns a CompletableFuture that completes when all processing
-     * (including cascaded work) finishes for this upload.
+     * Returns immediately with an {@link Upload} containing the upload ID and a future
+     * that completes when all processing finishes.
      *
      * @param name The name of the folder.
      * @param path The path within the folder.
      * @param data The JSON data to upload.
-     * @return A future that completes when all work for this upload is done.
+     * @return an Upload with the upload ID (safe to return to callers) and
+     *         a future (for callers that need to await completion).
      */
-    CompletableFuture<Void> upload(String name, String path, JqValue data);
+    Upload upload(String name, String path, JqValue data);
 
     /**
      * Recalculates all values in the folder by reprocessing each root value
--- a/src/main/java/io/hyperfoil/tools/h5m/rest/FolderResource.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/rest/FolderResource.java
@@ -18,11 +18,8 @@ import org.eclipse.microprofile.openapi.annotations.Operation;
 import org.eclipse.microprofile.openapi.annotations.parameters.Parameter;
 import org.eclipse.microprofile.openapi.annotations.tags.Tag;
 
-import io.smallrye.common.annotation.Blocking;
-
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.CompletionStage;
 
 @Path("/api/folder")
 @Produces(MediaType.APPLICATION_JSON)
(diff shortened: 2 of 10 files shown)

### CODE REVIEW

File: src/main/java/io/hyperfoil/tools/h5m/api/svc/FolderServiceInterface.java [hunk 1]
```diff
@@ -98,6 +99,14 @@ public interface FolderServiceInterface {
      */
     List<FolderSummary> getDashboardSummaries();
 
+    /**
+     * @param name Name of the folder
+     * @param path Path within the folder
+     * @param data The uploaded Json payload
+     * @return An upload Id in response
+     */
+    UploadHandle uploadGetId(String name, String path, JqValue data);
```
stalep: why not change upload to use the new return type?

src/main/java/io/hyperfoil/tools/h5m/api/UploadHandle.java
```diff
@@ -0,0 +1,17 @@
+package io.hyperfoil.tools.h5m.api;
+
+import jakarta.json.bind.annotation.JsonbTransient;
+
+import java.util.concurrent.CompletableFuture;
+
+public class UploadHandle {
```
stalep: I think we could reuse the similar name pattern we did for recalculate. Can we rename this UploadStatus?

src/main/java/io/hyperfoil/tools/h5m/rest/FolderResource.java
```diff
@@ -89,13 +87,15 @@ public long deleteFolder(@PathParam("name") String name) {
     @POST
     @Path("{name}/upload")
     @Authenticated
-    @Blocking
-    @Operation(description = "Upload JSON data to a folder. Returns when all processing completes.")
-    public CompletionStage<Void> upload(
+    @Operation(description = "Upload JSON data to a folder. Returns immediately with an uploadId.")
+    public UploadHandle upload(
```
stalep: Instead of returning the Upload object here, how about only return a long/Long and do `return upload(..).uploadId;` ?

src/main/java/io/hyperfoil/tools/h5m/api/UploadHandle.java
Arjun-Parmani: UploadStatus would imply that we are sending the status of the upload.  Whereas here we are only dealing with upload id. I think renaming could cause confusion.

src/main/java/io/hyperfoil/tools/h5m/api/svc/FolderServiceInterface.java
Arjun-Parmani: That will then lead to the existing methods that rely on CompleteableFuture<Void> to break .If some scenarios require to wait until the process is completed. furthermore if the return type of upload is changed, then all those methods specifically for ex LoadLegacyRuns which has a batchFuture() needs to be tweaked/altered as well.

src/main/java/io/hyperfoil/tools/h5m/rest/FolderResource.java
Arjun-Parmani: Well yes this can be updated to long. However the reasons to return an object is to maintain consistency with recalculate which returns object of RecalculationStatus and not just a raw status.

src/main/java/io/hyperfoil/tools/h5m/api/svc/FolderServiceInterface.java
stalep: Hm, but the existing methods can just call the CompletableFuture in UploadHandle?

Arjun-Parmani: Yes they can.

PR #165, opened 2026-07-02
disable processing resume for cli and add processing commands

willr3:


src/main/resources/application.properties
```diff
@@ -49,7 +49,7 @@ quarkus.banner.enabled=false
 # CLI mode: disable HTTP server and default to SQLite (works out of the box)
 # Override with -Dquarkus.datasource.db-kind=postgresql -Dquarkus.datasource.jdbc.url=jdbc:postgresql://host:port/db
 %cli.quarkus.http.host-enabled=false
-%cli.quarkus.datasource.db-kind=sqlite
+%cli.quarkus.datasource.db-kind=postgresql
```
stalep: Did you mean to change the default on the cli to use postgres?

willr3: Nope, good catch. I dream of a future where a single cli build works with either db :)

willr3: I fixed it in a second commit. If you find this suitable to merge please squash as I cannot do that in the mobile app

PR #166, opened 2026-07-03
Migrate CLI from picocli to aesh with REPL console mode

stalep:
Replace the picocli-based single-command CLI with an aesh interactive REPL (Read-Eval-Print Loop) shell. Users can now run multiple commands in a single session with tab completion, history, and folder context.

Entity-centric command structure:
  folder  — add, list, remove, upload, export, import, structure,
            recalculate, purge, values
  node    — add (jq/js/jsonata/sqlpath/sqlpathall/split/fixedthreshold/
            relativedifference/stddev), list, remove, update
  view    — list, show, create, remove, update (--add/--remove/--reorder)
  run     — list, show, upload
  notification — add, list, remove
  legacy  — load-tests, load-runs, verify
  admin   — create-user, create-team, add-member, list-users, list-teams,
            create-api-key, list-api-keys, revoke-api-key
  cd      — set active folder context
  upload  — top-level shortcut for folder upload

Key features:
- Folder context: 'cd myFolder' sets context, subsequent commands omit --to/--from
- Tab completion for commands, folder names, and node names
- acceptNameWithoutDashes on all options (picocli-compatible syntax)
- @Arguments for positional args matching original CLI syntax
- Default limit of 50 on list commands to prevent OOM
- Ctrl-C interrupt support (InterruptedException on all commands)
- generateHelp on all commands (--help shows usage)
- Views: create views, add/remove node columns, show pivoted data
- Runs: list uploads, show raw JSON data, per-node value inspection
- H5mCommandRegistryFactory filters top-level commands for clean tab completion
- Graceful startup recovery on SQLite (handles missing tables)
- Upload command awaits CompletableFuture completion before returning
- Uses aesh Table utility for formatted output
- NodeNameCompleter for node name tab completion in view/run commands


barreiro (changes requested):
great work. looks good overall.

many of comments that I have are around the usage of the API package. I'm fine if you decide not to address all of them now (just mark them as solved). there are other places where the API needs some more cleaning anyway.

barreiro (approved):

PR #171, opened 2026-07-03
Prevent concurrent recalculations and node updates during active reca…

stalep:
…lculation (issue #169)

Guard against race conditions caused by concurrent recalculations:

ProcessingService.doRecalculate():
- Added check for in-flight RECALCULATE and RECALCULATE_NODE trackers for the same folder. Rejects with IllegalStateException if another recalculation is already in progress.
- Recovery path bypasses the guard via isRecovery flag, since recovery replaces the crashed tracker with a new one.

NodeResource.update():
- Rejects node updates (name/operation changes) while a recalculation is active for the node's folder. Returns 400 Bad Request.
- Prevents the race where a node operation change during recalculation causes workers to read inconsistent operations.

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/api/RecalculationStatus.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/api/RecalculationStatus.java
@@ -16,5 +16,5 @@ public record RecalculationStatus(
         String error,
         long durationMs
 ) {
-    public enum State { RUNNING, COMPLETED, FAILED }
+    public enum State { RUNNING, COMPLETED, FAILED, CANCELLED }
 }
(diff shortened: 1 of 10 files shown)


src/main/java/io/hyperfoil/tools/h5m/rest/NodeResource.java
```diff
@@ -86,6 +86,23 @@ public NodeUpdateResponse update(@PathParam("id") Long id, NodeUpdateRequest req
         if (existing == null) {
             throw new NotFoundException("Node not found: " + id);
         }
+
+        // Reject node updates while a recalculation is in progress for this folder.
+        // Updating a node's operation during recalculation causes race conditions:
+        // in-flight workers may read the old or new operation depending on timing.
```
Arjun-Parmani: NodeResource.java has a guard which doesn't allow update to the node's operation and name while recalculation is in progress. Worth adding a test case for it in RestEndpointTest.java which calls PUT /api/node/{id} while a recalculation is tracker is active and asserts to HTTP 400 when trying to update node op/name?

stalep: Thanks, added that!

PR #177, opened 2026-07-05
Issue 172 batch fetch

stalep:


src/main/java/io/hyperfoil/tools/h5m/cli/LoadLegacyRuns.java
```diff
@@ -36,7 +37,7 @@ public class LoadLegacyRuns implements Callable<Integer> {
     @CommandLine.Option(names = {"username"}, description = "legacy db username", defaultValue = "quarkus") String username;
     @CommandLine.Option(names = {"password"}, description = "legacy db password", defaultValue = "quarkus") String password;
     @CommandLine.Option(names = {"url"}, description = "legacy connection url",defaultValue = "jdbc:postgresql://0.0.0.0:") String url;
-    @CommandLine.Option(names = {"testId"}, description = "specify which test to load. Loads all if unspecified" ) Long testId;
+    @CommandLine.Option(names = {"testId"}, description = "specify which test(s) to load. Comma-separated for multiple. Loads all if unspecified" ) String testId;
```
willr3: I believe picocli supports `@CommandLine.Option` on a `List` so that we don't have to maintain custom string parsing.

willr3 (commented):
Would you please check if `List` works for `testId`?

src/main/java/io/hyperfoil/tools/h5m/cli/LoadLegacyRuns.java
stalep: yup, fixed

stalep:
> Would you please check if `List` works for `testId`?

Works, updated the pr  :)

PR #179, opened 2026-07-05
Reduce O(n²) work queue sort overhead during bulk imports (issue #174)

stalep:
Three changes to reduce the allocation pressure from WorkQueue sorting:

1. Replace streams with loops in Work.dependsOn()
   - The nested activeNodes.stream().anyMatch() calls created ReferencePipeline, Spliterator, and MatchSink objects on every call.  With n² calls during sort, this generated up to 2.4 GB of transient allocations per 10-second window.
   - Extracted anyNodeDependsOn() and anyNodeDependsOnAny() helpers using simple for-each loops.

2. Skip sort when no node-level dependencies exist in queue
   - Maintain a Set<Long> of node IDs currently queued.
   - Before calling KahnDagSort, check if any queued work item's activeNode has an ancestor in the queued node ID set using NodeEntity.hasAncestor() (O(1) lookup via ancestor cache).
   - If no dependencies found, skip the entire O(n²) sort.
   - Added NodeEntity.hasAncestor(Long) for direct ID-based lookup.

3. Add visited set to ValueEntity.dependsOn() (correctness fix)
   - The BFS had no visited set, vulnerable to infinite loops on diamond-shaped value DAGs and exponential blowup on shared ancestors.

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/entity/ValueEntity.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/entity/ValueEntity.java
@@ -135,9 +135,13 @@ public class ValueEntity extends PanacheEntity {
     public boolean dependsOn(ValueEntity source){
         if(source == null) return false;
         Queue<ValueEntity> queue = new ArrayDeque<>(sources);
+        Set<Long> visited = new HashSet<>();
         boolean result = false;
         while(!queue.isEmpty() && !result){
             ValueEntity value = queue.poll();
+            if (value.id != null && !visited.add(value.id)) {
+                continue; // already visited — avoids redundant traversal of shared ancestors in diamond DAGs
+            }
             result = value.equals(source);
             if(!result){
                 queue.addAll(value.sources);
--- a/src/main/java/io/hyperfoil/tools/h5m/queue/WorkQueue.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/queue/WorkQueue.java
@@ -178,8 +178,17 @@ public class WorkQueue implements BlockingQueue<Runnable> {
         }
         return Collections.emptyList();
     }
+    // Tracks whether any queued Work depends on another queued Work.
+    // Set to true when addWorks detects a dependency. Reset after sort.
+    // When false, sort() is skipped (no dependencies to order).
+    private boolean dependenciesExist = false;
+
     private void sort(){
+        if (!dependenciesExist) {
+            return; // no dependencies between queued items — skip O(n²) sort
+        }
         runnables = KahnDagSort.sort(runnables,this::getRequiredPrecedingRunnables);
+        dependenciesExist = false;
     }
     public Collection<Work> addWorks(Collection<Work> works){
 //        putLock.lock();
@@ -199,6 +208,19 @@ public class WorkQueue implements BlockingQueue<Runnable> {
                 assert isPending(w);
             }).toList();
             if (!acceptedWork.isEmpty()) {
+                // Check if any new Work depends on any existing queued Work.
+                // Uses Work.dependsOn() as the single source of truth for dependency.
+                if (!dependenciesExist) {
+                    outer:
+                    for (Work newWork : acceptedWork) {
+                        for (Runnable r : runnables) {
+                            if (r instanceof Work existing && r != newWork && newWork.dependsOn(existing)) {
+                                dependenciesExist = true;
+                                break outer;
+                            }
+                        }
+                    }
+                }
                 sort();
                 if (wasEmpty) {
                     signalNotEmpty();


stalep:
Testing this change against the boot-time-verbose test showed a 2-3gb reduction in memory allocation.

src/main/java/io/hyperfoil/tools/h5m/queue/WorkQueue.java
```diff
@@ -178,8 +197,38 @@ public List<Runnable> getRequiredPrecedingRunnables(Runnable runnable){
...
+     * dependency.  Uses the pre-computed ancestorCache on NodeEntity for
+     * O(1) lookups per node, making this O(n × |queuedNodeIds|) instead
+     * of the O(n²) full dependsOn scan.
+     */
+    private boolean hasNodeDependenciesInQueue() {
+        if (queuedNodeIds.isEmpty()) {
+            return false;
+        }
+        for (Runnable r : runnables) {
+            if (r instanceof Work w && w.activeNodes != null) {
```
willr3: I think this is an abstraction violation where this code is copying part of the implementation details of `Work.dependsOn` and thus will break if we make changes to `Work.dependsOn` without coordinating those changes here. I think a more sustainable change would be to modify `Work` so that there is a method `hasNodeDependency` that is used by both `Work.dependsOn` and `WorkQueue`.

src/main/java/io/hyperfoil/tools/h5m/queue/WorkQueue.java
```diff
@@ -87,26 +98,32 @@ private void signalNotEmpty(){
...
-            }
-            for (int i = 0, size = runnables.size(); i < size; i++) {
-                Runnable r = runnables.get(i);
-                if (r instanceof Work wr && work.dependsOn(wr)) {
-                    return 1;
+        if(runnable instanceof Work work && work.activeNodes != null){
+            if (!activeNodeIds.isEmpty()) {
+                for (NodeEntity n : work.activeNodes) {
+                    for (Long activeId : activeNodeIds) {
+                        if (n.hasAncestor(activeId)) {
```
willr3: I believe this simplification of the algorithm is invalid if the work is `cumulative=true`. This is only taking activeNode dependency into consideration. I also think this change makes it harder to maintain the project because the logic for work dependency is being spread from a single `Work.dependsOn` method and into multiple classes and methods. Do you think copying the logic into multiple classes and methods negatively impacts project maintainability?

willr3 (changes requested):
I think merging this would make the project harder to maintain because it is taking the implementation details of `Work.dependsOn` and copying them to multiple other locations in the code. That means changing how dependencies are managed would require coordination multiple classes.

stalep:
I refactored two methods on Work that serve as the single source of truth for node-level dependency checking:

- `Work.hasNodeDependency(Work other)`, returns true if any of this Work's active nodes depends on any of the other Work's active nodes. Used by both `dependsOn()` and` WorkQueue.getScore()`.
- `Work.hasNodeDependencyOnAny(Set<Long> nodeIds)`, returns true if any of this Work's active nodes has an ancestor in the given set of node IDs. Used by` WorkQueue.hasNodeDependenciesInQueue()` for the sort-skip optimization.

`Work.dependsOn()` now calls `hasNodeDependency()` for its node-level check instead of inline loops. WorkQueue delegates to these methods instead of re-implementing the check.

willr3:
I think this change breaks some of the functionality for `WorkQueue` that worked because `WorkQueue` used `Work.dependsOn` and `dependsOn` had tests to ensure correctness. I've added an example test `WorkQueue.poll_return_child_node_with_different_sourceValue`  that passes in `main` but fails for this PR to illustrate why this PR cannot be merged in the current state.

willr3 (changes requested):
I made an earlier commit but if you rebase I added `WorkQueue.poll_return_child_node_with_different_sourceValue` to illustrate where this breaks WorkQueue

src/main/java/io/hyperfoil/tools/h5m/entity/ValueEntity.java
```diff
@@ -134,15 +134,19 @@ public boolean equals(Object o){
...
-        boolean result = false;
-        while(!queue.isEmpty() && !result){
+        while(!queue.isEmpty()){
             ValueEntity value = queue.poll();
-            result = value.equals(source);
-            if(!result){
+            if(value.equals(source)){
+                return true;
+            }
+            // Track visited nodes to avoid infinite loops on diamond-shaped
```
willr3: How does a diamond-shaped DAG create an infinite loop?

src/main/java/io/hyperfoil/tools/h5m/queue/WorkQueue.java
```diff
@@ -87,26 +93,28 @@ private void signalNotEmpty(){
...
     }
 
+    /**
+     * Returns 0 if the item at the given index is unblocked and ready to run,
+     * or a positive value if it must wait.
+     *
+     * A Work item is blocked if it depends on any currently executing (active)
+     * Work item. Uses {@link Work#dependsOn(Work)} which checks both node-level
+     * dependency AND source value overlap — a child node with a different source
+     * value is NOT blocked by an active parent work item.
```
willr3: Removing the `runnable` scan for dependencies means the `getScore` method will only be correct up to the first unblocked entry in queue. After that point `getScore` would not be reliable.

src/main/java/io/hyperfoil/tools/h5m/queue/WorkQueue.java
```diff
@@ -560,4 +597,30 @@ public int drainTo(Collection<? super Runnable> c, int maxElements) {
...
+        if (work.activeNodes != null) {
+            for (NodeEntity n : work.activeNodes) {
+                if (n.getId() != null) {
+                    queuedNodeIds.add(n.getId());
+                }
+            }
+        }
+    }
+
+    private void removeNodeIds(Work work) {
```
willr3: I believe rebuilding the `queuedNodeIds` each time a Work is removed from the queue is O(n^2)

src/main/java/io/hyperfoil/tools/h5m/entity/work/Work.java
```diff
@@ -60,28 +60,51 @@ public void setActiveNodes(Set<NodeEntity> activeNodes) {
...
+        }
+        return false;
+    }
+
+    /**
+     * Returns true if any of this Work's active nodes has an ancestor in the
+     * given set of node IDs. Used by WorkQueue to quickly check if any pending
+     * work items have node-level dependencies without iterating all Work pairs.
+     */
+    public boolean hasNodeDependencyOnAny(Set<Long> nodeIds) {
```
willr3: My concern is that this exposes an implementation detail of Work.dependsOn without being part of how `dependsOn` is implemented. That creates a parallel implementation that has to be maintained in synchronization with `dependsOn` and if they fall out of sync then Value calculation would be jeopardized.

willr3 (changes requested):
I'm concerned that `hasNodeDependencyOnAny` is a parallel implementation that is not part of how `Work.dependsOn` functions and therefore is an increase in the maintenance burden for the core queueing functionality.

src/main/java/io/hyperfoil/tools/h5m/entity/ValueEntity.java
stalep: updated the comment

PR #180, opened 2026-07-05
Replace Work.sourceValues List<ValueEntity> with List<Long> IDs (issu…

stalep:
…e #178)

Work items no longer carry full ValueEntity objects through the queue. Instead they carry only value IDs — full entities are loaded in WorkService.execute() via em.find() (which hits the 2LC cache).

This eliminates the primary remaining source of memory retention during bulk imports: queued Work items carrying multi-megabyte parsed JSON trees (JqValue data field) through the work queue.

Changes:
- Work.sourceValues -> Work.sourceValueIds (List<Long>)
- Remove value-level dependency check from Work.dependsOn() — the node-level dependency check is sufficient for ordering, and the dedup logic in WorkService.execute() handles duplicate values. The removed check walked the ValueEntity.sources DAG which required full entity references.
- Simplify WorkService.execute() reload loop — always load by ID instead of checking Hibernate.isPropertyInitialized()
- Update all Work creation sites to pass value IDs: FolderService.upload(), ProcessingService, WorkService cascade
- Update Work.equals(), hashCode(), toString() for Long-based IDs
- Update WorkService.findTrackers() to iterate IDs directly

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/entity/work/Work.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/entity/work/Work.java
@@ -1,7 +1,6 @@
 package io.hyperfoil.tools.h5m.entity.work;
 
 import io.hyperfoil.tools.h5m.entity.NodeEntity;
-import io.hyperfoil.tools.h5m.entity.ValueEntity;
 import io.hyperfoil.tools.h5m.entity.node.EDivisive;
 import io.hyperfoil.tools.h5m.entity.node.RelativeDifference;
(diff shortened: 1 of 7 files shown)


willr3:
Does this change continue to support `Work` with `souceValues` but no `activeNodes`?

stalep:
> Does this change continue to support `Work` with `souceValues` but no `activeNodes`?

Yes,  the behavior is identical to before:

- `dependsOn()` returns false immediately when `this.activeNodes` is null or empty
- `execute()` exits early when activeNodes is empty
- The `sourceValueIds` list is preserved on the Work item

PR #185, opened 2026-07-06
WorkQueue.addAll() iterates input collection and returns correct boolean

Arjun-Parmani:
addAll() had a bug where the loop iterated over runnables (the existing internal queue) instead of the input collection c. This meant calling addAll() never actually enqueued any of the provided work items, silently discarding them. The method also always returned false regardless if the work is added to the pendingwork() or not.

**Changes:**
  - Fix loop to iterate over the input collection c instead of runnables
  - Track whether any item was actually added and return  @true
  - Guard sort() behind the added flag — no point re-sorting when nothing changed

  **Tests added (WorkQueueTest):**
  - addAll_adds_new_items — happy path, all new items enqueued, returns true
  - addAll_rejects_all_duplicates — items already pending are skipped, returns false
  - addAll_check_and_adds_only_new — mixed input; duplicate is skipped, new item is enqueued, size delta is exactly 1
  
 Closes #184

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/queue/WorkQueue.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/queue/WorkQueue.java
@@ -435,22 +435,30 @@ public class WorkQueue implements BlockingQueue<Runnable> {
     @Override
     public boolean addAll(Collection<? extends Runnable> c) {
         fullyLock();
+        boolean added = false;
+        boolean wasEmpty = runnables.isEmpty();
         try {
-            for(Runnable r : runnables){
-                if( r instanceof Work work && isPending(work)){
+            for (Runnable r : c) {
+                if (r instanceof Work work && isPending(work)) {
                     //do not add something already in queue
-                }else{
-                    if( r instanceof Work work){
+                } else {
+                    if (r instanceof Work work) {
                         pendingWork.add(work);
                     }
                     runnables.add(r);
+                    added = true;
                 }
             }
-            sort();
-        }finally {
+            if (added) {
+                sort();
+                if(wasEmpty){
+                    signalNotEmpty();
+                }
+            }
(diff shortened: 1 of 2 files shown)


src/main/java/io/hyperfoil/tools/h5m/queue/WorkQueue.java
```diff
@@ -435,22 +435,26 @@ public boolean containsAll(Collection<?> c) {
     @Override
     public boolean addAll(Collection<? extends Runnable> c) {
         fullyLock();
+        boolean added = false;
         try {
-            for(Runnable r : runnables){
```
stalep, line 439: If here we add:
boolean wasEmpty = runnables.isEmpty();

src/main/java/io/hyperfoil/tools/h5m/queue/WorkQueue.java
```diff
@@ -435,22 +435,26 @@ public boolean containsAll(Collection<?> c) {
...
                         pendingWork.add(work);
                     }
                     runnables.add(r);
+                    added = true;
                 }
             }
-            sort();
-        }finally {
+            if (added) {
+                sort();
```
stalep, line 453: We can here add;
```
if (wasEmpty)
  signalNotEmpty();
```

src/test/java/io/hyperfoil/tools/h5m/queue/WorkQueueTest.java
```diff
@@ -15,6 +15,8 @@
 import org.junit.jupiter.api.Test;
 
 import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
```
stalep: I do not think these two are needed?

PR #192, opened 2026-07-07
Fix import and verifier: disconnected nodes, naming, per-upload scopi…

stalep:
…ng (issue #191)

Import fixes (LoadLegacyTests):
- Add group.addNode(rtrn) for JS combination function nodes
- Handle JS functions with default parameters via countNonDefaultParams()
- Strip JSDoc/line comments before parsing function signatures
- Skip labels with zero extractors (placeholder/stub labels)
- Return reused extractor directly for identity labels instead of creating solo=>solo wrapper (removes !reusedNode gate)
- Rename single-variant nodes back to original label name after multi-schema dedup (fixes User0→User, UUID0→UUID, etc.)

Verifier fixes (VerifyLegacy):
- Scope h5m values to specific upload via recursive CTE from root value
- Get root values via node_group.root_id join (root nodes have null group_id)
- Match values by node name rather than idx offset
- Add fallback name matching: try with numeric suffix (User→User0), lowercase (User→user), and snake_case (Run ID→run_id)
- Show root value ID in output, skip runs with no h5m upload

Results for boot-time-verbose (5 runs):
  Before: 24.6% match (49/199), 2 labels at 100% After:  66.8% match (133/199), 27 labels at 100%

Remaining failures (35 missing, 31 mismatch) are from:
- Multi-schema data path mismatches (jsonpath for one schema version doesn't match data from another version)
- Stub JS functions returning constants ("Need to collect")
- SUT combination function with incomplete source resolution
- Change detection nodes needing more data points

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/cli/LoadLegacyTests.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/cli/LoadLegacyTests.java
@@ -4,6 +4,7 @@ import io.hyperfoil.tools.jjq.value.JqArray;
 import io.hyperfoil.tools.jjq.value.JqObject;
 import io.hyperfoil.tools.jjq.value.JqValue;
 import io.hyperfoil.tools.jjq.value.JqValues;
+import io.hyperfoil.tools.h5m.api.EphemeralMode;
 import io.agroal.api.AgroalDataSource;
 import io.agroal.api.configuration.supplier.AgroalPropertiesReader;
 import io.hyperfoil.tools.h5m.entity.FolderEntity;
@@ -101,6 +102,10 @@ public class LoadLegacyTests implements Callable<Integer> {
         public void tagNodeAsLabel(Label label,NodeEntity node){
             labelToNodes.put(label,node);
             nodeToLabel.put(node,label);
+            // Label-equivalent nodes should retain their data after processing.
+            // Without KEEP, the ephemeral system nullifies them as "intermediate"
+            // nodes, but in Horreum these are the final label values.
+            node.ephemeral = EphemeralMode.KEEP;
         }
         public void renameNode(NodeEntity node,String oldName){
             nodesByName.remove(oldName,node);
@@ -230,15 +235,17 @@ public class LoadLegacyTests implements Callable<Integer> {
     //id,name,labels,calculation
     public record Variable(long id,String name,List<String> labels,String calculation){};
 
-    public NodeEntity createNodesFromLabel(Label label, NodeEntity source, NodeGroupEntity group, NodeTracking nodeTracking, Set<String> usedNames){
+    public NodeEntity createNodesFromLabel(Label label, NodeEntity source, NodeGroupEntity group, NodeTracking nodeTracking, Set<String> usedNames, boolean multiSchema){
+        // Fix A: Skip labels with no extractors — these are placeholder/stub labels
+        // from schemas like rhivos-perf-comprehensive:02 that contribute no data
+        if (label.extractors.isEmpty()) {
+            return null;
+        }
         NodeEntity rtrn = null;
         HashedLists<String,NodeEntity> labelNodesByName = new HashedLists<>();
         boolean reusedNode = false;
         for(Extractor extractor : label.extractors) {
             String extractorName = extractor.name;
-            if(extractorName.equals(label.name) || usedNames.contains(extractorName)){
-                extractorName = extractor.name;
-            }
 
             // Convert jsonpath to jq — sqlall (isArray) wraps in [...] to collect all matches
             String jqOperation = extractor.isArray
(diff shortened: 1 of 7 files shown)


src/main/java/io/hyperfoil/tools/h5m/cli/LoadLegacyTests.java
```diff
@@ -327,11 +346,58 @@ public NodeEntity createNodesFromLabel(Label label, NodeEntity source, NodeGroup
...
         }
         return rtrn;
     }
 
+    /**
+     * Counts the number of JS function parameters that do NOT have default values.
+     * For example, {@code (arr, usePopulation = false)} has 1 non-default parameter.
+     * Parameters with defaults are JS-level config, not data sources from the pipeline.
+     */
+    static int countNonDefaultParams(String function) {
```
willr3: This should be part of JsNode. This type of analysis will be needed each time a source node name changes.

src/main/java/io/hyperfoil/tools/h5m/cli/LoadLegacyTests.java
```diff
@@ -286,7 +296,21 @@ public NodeEntity createNodesFromLabel(Label label, NodeEntity source, NodeGroup
...
             if (rtrn == null) {
                 List<String> params = JsNode.getParameterNames(label.function);
-                if(params.size()==1 && label.extractors.size() > 1) {
+                // Count parameters without default values — these are the actual data sources.
+                // Parameters with defaults (e.g., usePopulation = false) are JS-level config,
+                // not data sources from the pipeline.  In Horreum, the mapping is positional:
+                // first extractor → first JS parameter, second extractor → second JS parameter, etc.
+                // Extra parameters with defaults are NOT data sources.
+                int dataParamCount = countNonDefaultParams(label.function);
+                if (dataParamCount > 0 && dataParamCount < params.size()) {
```
willr3: I believe this should be part of the `JsNode.parse` functionality and not part of `LoadLegacyTests`. The responsability of `JsNode.parse` is to create a valid JsNode when given valid input, having this edge case supported here breaks the abstraction of the JsNode.parse method.

willr3 (changes requested):
I think the fixes proposed here in LoadLegacyTests belong in JsNode to maintain the abstraction and isolation of responsibilities between the classes.

stalep:
> I think the fixes proposed here in LoadLegacyTests belong in JsNode to maintain the abstraction and isolation of responsibilities between the classes.

Yep, I agree. Moved them now.

src/main/java/io/hyperfoil/tools/h5m/entity/node/JsNode.java
```diff
@@ -163,6 +174,57 @@ public static boolean isNullEmptyOrIdentityFunction(String input){
             ;
     }
 
+    /**
+     * Returns the set of parameter names that have default values in the function
+     * signature. For example, {@code (arr, usePopulation = false)} returns
+     * {@code {"usePopulation"}}. These parameters are JS-level config, not data
+     * sources from the pipeline.
+     */
+    public static Set<String> getDefaultParameterNames(String input) {
```
willr3: This function looks like it was a copy and paste of `getParameterNames` with modifications after the paste. I think it would be better for the project to avoid duplicating several dozen lines of code that would have to be maintained in parallel. Perhaps add a control parameter to getParameterNames or abstract the common code to a helper method?

src/main/java/io/hyperfoil/tools/h5m/cli/LoadLegacyTests.java
```diff
@@ -288,7 +298,14 @@ public NodeEntity createNodesFromLabel(Label label, NodeEntity source, NodeGroup
             rtrn = JsNode.parse(label.name, function, labelNodesByName::get);
             if (rtrn == null) {
```
willr3: Do we need this logic if a user creates a javascript function outside of loading the legacy database? If so then the code should be in a re-usable location.  I'm honestly not convinced we need the `combiner` node. Do you have any use cases where that node was necessary for correctness?

willr3 (changes requested):
There are still issues that I believe we should discuss / resolve before we merge this change.

src/main/java/io/hyperfoil/tools/h5m/cli/LoadLegacyTests.java
stalep: The combiner is needed for correctness. When multiple schema variants exist, the import appends numeric suffixes to deduplicate extractor nodes (e.g., rhivos_target -> rhivos_target1). Without the combiner, createParameters() would pass {"rhivos_target1": ...} to the JS function, but the function accesses value["rhivos_target"] (no suffix), the lookup would return undefined.

The combiner preserves the original Horreum extractor names as keys in its jq expression, so the js function's property accesses work regardless of how h5m names the nodes internally.

PR #195, opened 2026-07-08
Folder CRUD - Add folder CRUD operations to Dashboard UI

Arjun-Parmani:
Summary:
  - Added a Create Folder button on the Dashboard page that opens a modal allowing users to enter a folder name and save it to the database via POST /api/folder/{name}
  - Added a Delete button on each folder row that prompts for confirmation before calling DELETE /api/folder/{name}
  - Both operations refresh the dashboard table automatically on success without a full page reload
  - Duplicate folder name error is handled with a user-friendly message instead of a raw 500 error
  
To Test:
Enter into dev mode - mvn quarkus:dev - open http://localhost:8080/
  - Create a new folder from the UI and verify it appears in the table and the database
  - Try creating a folder with the same name and verify the friendly error message appears
  - Delete a folder and verify it disappears from the table and the database
  - Verify clicking Cancel on both modals closes them without making any changes

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/svc/FolderService.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/svc/FolderService.java
@@ -29,6 +29,7 @@ import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
 import jakarta.persistence.EntityManager;
 import jakarta.transaction.Transactional;
+import jakarta.ws.rs.WebApplicationException;
 import org.hibernate.query.NativeQuery;
 
 import java.io.IOException;
@@ -72,6 +73,9 @@ public class FolderService implements FolderServiceInterface {
     @Override
     @Transactional
     public long create(String name){
+        if(FolderEntity.find("name",name).firstResult() !=null){
+            throw new WebApplicationException("Folder already exists:" + name, 409);
+        }
         FolderEntity entity = new FolderEntity();
         entity.name = name;
         entity.group = new NodeGroupEntity(name); //TODO do we auto-create a nodeGroup?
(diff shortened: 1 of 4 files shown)


stalep:
Looks ok, but I think we need to improve a few things;
- The "Create Folder" button need to show if there are no folders as well
- If the user triggers an error, closes the modal and reopens it the error message persist. I think onClose handler should clear the state?
- Add onError handler for delete failure
- Replace hardcoded colors with Carbon defaults also replace window.confirm() with Carbon modal I think

I also think 500 == duplicate name is brittle. There might be other issues which could cause a 500. I think we can use 409 for duplicate name and 500 for other errors?

src/main/webui/src/app/pages/DashboardPage.tsx
```diff
@@ -81,8 +90,27 @@ const SummaryCards = ({ summaries }: { summaries: FolderSummary[] }) => {
...
+    ...deleteFolderMutation(),
+    onSuccess: () => {
+      void queryClient.invalidateQueries();
+      setConfirmFolder(null);
+    },
+    onError: (e: AxiosError<Error>) => {
+      if (e.response?.status === 500) {
+              setDeleteError('This folder contains nodes or uploaded data and cannot be deleted. Please remove all nodes and data first.');
+            } else {
+              setDeleteError(e.message ?? 'Failed to create folder');
```
stalep: replace create with delete :)

src/main/java/io/hyperfoil/tools/h5m/svc/FolderService.java
```diff
@@ -73,6 +74,9 @@ private static LocalDateTime toLocalDateTime(Object value) {
     @Transactional
     public long create(String name){
         FolderEntity entity = new FolderEntity();
+        if(entity.find("name",name).firstResult() !=null){
```
stalep: Use FolderEntity.find instead of entity.find and move the initialization of entity after the if bracket I think.

Issue #154, opened 2026-06-24
Replace SQL/SQLAll jsonpath nodes with jq nodes


stalep:
## Motivation

The `sql` and `sqlall` node types execute PostgreSQL jsonpath expressions (`jsonb_path_query_first`/`jsonb_path_query_array`) via native SQL queries. Since #150, all JSON data lives in memory as `JqValue` during pipeline evaluation — the database round-trip is unnecessary. The same expressions can be evaluated directly by the jq VM using the existing `jsonpathToJq()` converter.

Removing these node types:
- **Eliminates a DB round-trip** per sql/sqlall node evaluation (native query + text parsing)
- **Removes database-specific code** (the `switch(dbKind)` dispatch for PostgreSQL vs SQLite)
- **Simplifies the codebase** — two entity classes, two CLI commands, and ~100 lines of SQL generation/parsing code
- **Makes all expression evaluation consistent** — everything goes through the jq VM

h5m is not yet in production, so there is no live database to migrate. The only migration path needed is `LoadLegacyTests` (which creates sql/sqlall nodes from legacy Horreum extractors).

## Conversion

Every jsonpath expression has a direct jq equivalent. The existing `NodeService.jsonpathToJq()` converter already handles the translation:

| jsonpath (sql) | jq |
|---|---|
| `$.foo.bar` | `.foo.bar` |
| `$.metadata.user` | `.metadata.user` |
| `$.results."CoreMark-PRO".SingleCore` | `.results."CoreMark-PRO".SingleCore` |

| jsonpath (sqlall) | jq |
|---|---|
| `$.results[*].iops` | `[.results[].iops]` |
| `$.stressng_workload[*].test_results` | `[.stressng_workload[].test_results]` |
| `$.results.jobs[*].read.iops` | `[.results.jobs[].read.iops]` |

The key difference: `sqlall` uses `jsonb_path_query_array` which returns all matches as an array. The jq equivalent wraps the expression in `[...]` to collect multiple outputs into an array.

### Phase 1: Convert test data in-place

- Convert all `"type": "sql"` and `"type": "sqlall"` entries in `rhivos/nodes.json` to `"type": "jq"` with converted operations
- Update `LoadLegacyTests` to create `JqNode` instead of `SqlJsonpathNode`/`SqlJsonpathAllNode`, using `jsonpathToJq()` for conversion (array extractors get `[...]` wrapper)

### Phase 2: Remove node types and entities

**Delete files:**
[...]

PR #196, opened 2026-07-10
Remove leftover SQL/SQLAll test references (issue #154 follow-up)

stalep:
The SQL/SQLAll node types were removed in issue #154, but some test methods in NodeServiceTest kept the old 'calculateSql*' names despite already using JqNode internally.

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/svc/FolderService.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/svc/FolderService.java
@@ -482,7 +482,7 @@ public class FolderService implements FolderServiceInterface {
         return builder.build();
     }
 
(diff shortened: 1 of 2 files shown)


willr3:
I agree with renaming the test methods to accurately reflect that they are testing `calculateJqValues` and not `calculateSqlJsonpathValues` but deleting the tests that verify `calculateJqValues` provides the expected output to support the `sqlall` use case is not something I agree with. Would you be willing to rename those test methods (and the ohter ones that are already being renamed) to indicate they are testing that jqNodes can replace the old Horreum sqlall functionality?

willr3 (changes requested):
Please do not delete the calculateSqlJsonpathValues methods. They should be renamed to indicate they validate calculateJqValues supports the expected behavior for nodes that were originally sql and sqlall

stalep:
I've renamed the tests and reinstated them.

PR #198, opened 2026-07-13
fix WorkQueue to reject RootNode as activeNode

willr3:


willr3:
I'm not sure how I managed to have the commit in the history of another PR but :shrug:

PR #199, opened 2026-07-13
Fix root node ephemeral

willr3:


src/test/java/io/hyperfoil/tools/h5m/queue/WorkQueueTest.java
```diff
@@ -27,6 +29,76 @@ public class WorkQueueTest extends FreshDb {
...
+        Work work1 = new Work(rootNode, Collections.emptyList(),List.of(rootValue1.id));
+        Work work2 = new Work(jqNode, Collections.emptyList(),List.of(rootValue1.id));
+        tm.commit();
+
+        WorkQueue q = new WorkQueue();
+
+        boolean added = q.addAll(List.of(work1,work2));
+        assertTrue(added,"at least one was added");
+        assertFalse(q.isPending(work1),"root should not be pending");
+        assertFalse(q.hasWork(work1),"root should not be queued");
```
Arjun-Parmani, line 100: The test indirectly proves that work2 is added in the queue. Does having a proper assert makes it more concrete? Something like assertTrue(q.isPending(work2),"jqNode is pending")? same for hasWork

willr3: `assertTrue(added)` ensures the return of `addAll` is `true` despite one of the values being rejected. We could add `assertTrue(q.hasWork(work2))`. I will update the test cases.

willr3: updated

stalep:
I think we should change `NodeEntity.ephemeral` to private as the no-op setEphemeral on RootNode isn't really protecting anything atm.

willr3:
> I think we should change `NodeEntity.ephemeral` to private as the no-op setEphemeral on RootNode isn't really protecting anything atm.

Would you please elaborate what you mean by saying the setEphemeral on RootNode isn't really protecting anything? 

I tested your statement by commenting out the method on RootNode and that causes `io.hyperfoil.tools.h5m.entity.NodeTest#root_setEphemeral` to fail.

My understanding of Panache is that it creates getters and setters for public member variables and converts variable references to use those methods. Providing a `setEphemeral` implementation on RootNode leads Panache to use that implementation instead of creating one because of the getter and setter naming convention. 

Were you able to create a test that demonstrates that `setEphemeral` on `RootNode` was not protecting the `ephemeral` value?

stalep:
Ah, I forgot that panache handles that, forget what I said!

PR #201, opened 2026-07-14
add Value filtering to ValueService.getGroupedValues

willr3:


src/test/java/io/hyperfoil/tools/h5m/svc/ValueServiceTest.java
```diff
@@ -1277,7 +1277,93 @@ public void getDirectDescendantValues() throws SystemException, NotSupportedExce
...
+                """
+                {
+                  "group": [
+                    { "a": 2, "b": 1, "c": "twoone"},
+                    { "a": 2, "b": 2, "c": "twotwo"},
+                  ]
+                }
+                """
+        ));
+        root1.persist();
```
stalep: I think you meant to persist root2 and not root1 again? - also there are trailing , in the json here and above.

stalep:
Other than the one comment, lgtm :+1:

src/test/java/io/hyperfoil/tools/h5m/svc/ValueServiceTest.java
willr3: Good catch, I corrected both in the PR branch.

PR #203, opened 2026-07-15
Nodes with multi parent now deleted

Arjun-Parmani:
Deleting intermediate nodes (JQ, JS, JSONATA, Fingerprint) from multi-source DAG pipelines failed with a 500 FK constraint violation. The root cause was that NodeService.delete() only removed outgoing edges (where the node is a parent), but left incoming edges (where the node is a child of its own sources) intact. When the DB tried to delete the node row, those node_edge.child_id references caused the constraint violation.

  **Changes**

1.   NodeService.delete()

  - Added EdgeQueries.deleteChildEdges() to remove incoming node_edge rows (edges from the node's parents to itself) before deleting the node. This is the primary fix for the FK violation.
  - Added deletion of all ValueEntity rows associated with the node before removing it, since value.node_id is a non-nullable FK. Without this, deleting any node that has accumulated values from prior uploads would fail at the DB level.

2.   NodeServiceTest.delete_does_not_cascade_to_shared_dependent()

  - Added the missing assertion that sourceA is null after deletion. The original test verified that sharedNode and sourceB survived, but never confirmed that sourceA itself was actually removed.

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
@@ -286,8 +286,16 @@ public class NodeService implements NodeServiceInterface {
                     delete(dependent.id);
                 }
             }
+
+            List<ValueEntity> nodeValues = ValueEntity.find("node.id", nodeId).list();
+            for (ValueEntity v : nodeValues) {
+                valueService.delete(v);
+            }
+            EdgeQueries.deleteChildEdges(em, "node_edge", node.id);
             // clean up edge rows where this node is a parent (inverse side not managed by JPA)
             EdgeQueries.deleteParentEdges(em, "node_edge", nodeId);
+            em.flush();
+            em.clear();
             NodeEntity.deleteById(nodeId);
         }
     }
--- a/src/test/java/io/hyperfoil/tools/h5m/svc/NodeServiceTest.java
+++ b/src/test/java/io/hyperfoil/tools/h5m/svc/NodeServiceTest.java
@@ -95,12 +95,61 @@ public class NodeServiceTest extends FreshDb {
         nodeService.delete(sourceA.id);
         tm.commit();
 
+        // sourceA should be deleted
+        NodeEntity nodeA = NodeEntity.findById(sourceA.id);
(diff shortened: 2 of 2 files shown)


src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
```diff
@@ -286,8 +286,16 @@ public void delete(Long nodeId){
...
+
+            List<ValueEntity> nodeValues = ValueEntity.find("node.id", nodeId).list();
+            for (ValueEntity v : nodeValues) {
+                valueService.delete(v);
+            }
+            EdgeQueries.deleteChildEdges(em, "node_edge", node.id);
             // clean up edge rows where this node is a parent (inverse side not managed by JPA)
             EdgeQueries.deleteParentEdges(em, "node_edge", nodeId);
+            em.flush();
+            em.clear();
```
stalep, line 298: Is calling em.clear() needed? (might be, I'm just curious :)

src/test/java/io/hyperfoil/tools/h5m/svc/NodeServiceTest.java
```diff
@@ -95,6 +95,9 @@ public void delete_does_not_cascade_to_shared_dependent() throws HeuristicRollba
         nodeService.delete(sourceA.id);
         tm.commit();
 
+        // sourceA should be deleted
+        NodeEntity NodeA = NodeEntity.findById(sourceA.id);
```
stalep: rename NodeA to nodeA

stalep (commented):
I think we should add some more tests that show what this fixes. Eg add a test where we delete a node that has multiple parents and delete a node that has associated values?

src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
Arjun-Parmani: Good catch! We delete the edges first using raw SQL, so Hibernate doesn't know about it and still has the old state in memory. Without em.clear(), it would undo our deletions before the node gets removed. Clearing it first prevents that.

PR #208, opened 2026-07-17
Add upload status tracking and per-upload change detection (issue #205)

stalep:
Adds REST endpoint for polling upload processing status and retrieving change detection results triggered by specific uploads.

New endpoint:
  GET /api/folder/upload/{uploadId}/status
  Returns UploadStatus: state (PROCESSING/COMPLETED/FAILED), durationMs,
  error message, and list of Change records with changeType enum.

Implementation:
- UploadService: in-memory bounded tracker map (max 1000 entries, 10-min retention after completion). Observes ChangeDetectedEvent to capture per-upload detection results (dispatch=true only).
- UploadStatus/Change DTOs in api package. Change.ChangeType enum maps from NodeType detection types.
- NodeType.toChangeType() converts detection node types to Change enum.
- ChangeDetectedEvent: added rootValueId and nodeType fields so the UploadService observer can attribute changes to specific uploads.
- FolderService.upload(): registers with UploadService for both the normal and empty-folder (no work) paths.


src/main/java/io/hyperfoil/tools/h5m/svc/UploadService.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    private void evictOldestCompleted() {
+        trackers.entrySet().stream()
+                .filter(e -> e.getValue().state != UploadStatus.State.PROCESSING)
+                .min((a, b) -> Long.compare(a.getValue().completedAt, b.getValue().completedAt))
+                .ifPresent(e -> trackers.remove(e.getKey()));
+    }
+
+    private static class Tracker {
```
Arjun-Parmani: Why Tracker instead of extending Upload.java?
UploadService.Tracker holds uploadId, startedAt, state, error, completedAt, and changes. But Upload.java already represents the same entity. the uploadId and the CompletableFuture<Void> that tracks completion of that exact upload.

What was the reasoning for introducing a separate Tracker inner class here rather than building on top of Upload.java? Was there a specific constraint that made reusing Upload impractical? - Just curious

stalep: The idea was that Upload is a rest response and short-lived. The tracker has a different lifecycle and concern. It's persisted for 10min and accumulates change detection results from the cdi events.
Also, I was thinking that not all uploads are interested in polling for changes....
Then again, the Tracker name is not great... - could be UploadMonitor?

src/test/java/io/hyperfoil/tools/h5m/rest/RestEndpointTest.java
```diff
@@ -1092,8 +1092,66 @@ public void multiple_uploads_return_unique_ids() {
...
+        // Allow a brief wait for async processing
+        try { Thread.sleep(500); } catch (InterruptedException e) { Thread.currentThread().interrupt(); }
 
+        given()
+                .when().get("/api/folder/upload/" + uploadId + "/status")
+                .then()
+                .statusCode(200)
+                .body("uploadId", equalTo(uploadId.intValue()))
+                .body("state", equalTo("COMPLETED"))
+                .body("changes", notNullValue());
```
Arjun-Parmani: The test has a node of type 'extract' but the test never checks the actual implementation. There is no detection node which has a threshold/boundary which highlights that the uploaded data has violated the threshold and a change is fired. The empty change list is deceiving here. I think there should be a test that covers this: a detection node that triggers a change and then change list ls returned to the user.

stalep: Thanks @Arjun-Parmani , I added more tests now.

src/main/java/io/hyperfoil/tools/h5m/svc/UploadService.java
```diff
@@ -0,0 +1,212 @@
...
+                    SELECT ve.child_id FROM value_edge ve WHERE ve.parent_id = :rootId
+                    UNION ALL
+                    SELECT ve.child_id FROM value_edge ve JOIN descendants d ON ve.parent_id = d.vid
+                )
+                SELECT v.id, v.node_id, n.name, n.type
+                FROM value v
+                JOIN node n ON v.node_id = n.id
+                JOIN descendants d ON v.id = d.vid
+                WHERE n.type IN """ + NodeService.DETECTION_NODES)
+                .setParameter("rootId", uploadId)
```
willr3: I believe it is an abstraction violation to create this functionality on `UploadService.` The `ValueService` is responsible for retrieving `Values` from persistence and we should strive to keep the code organized. It looks like this is trying to get the descendants of rootId that were created by a subset of node types. Pehraps we should add that to the `ValueService` API instead of breaking the db abstraction?

willr3 (commented):
I do not think we should have sql to fetch `Values` outside of the `ValueService`.

src/main/java/io/hyperfoil/tools/h5m/rest/ValueResource.java
```diff
@@ -14,8 +16,9 @@
 import org.eclipse.microprofile.openapi.annotations.tags.Tag;
 
 import java.util.List;
+import java.util.Optional;
 
-@Path("/api/value")
+@Path("/api")
```
willr3: value service should be `/api/value` not the top level `/api`?

src/main/java/io/hyperfoil/tools/h5m/rest/ValueResource.java
```diff
@@ -47,26 +54,73 @@ public JqValue getValueData(@PathParam("id") Long id) {
...
         return valueService.getNodeValues(nodeId);
     }
+
+    // ---- Processing status ----
+
+    @GET
+    @Path("processing/{id}")
+    @PermitAll
+    @Operation(description = "Get the processing status of an upload. Returns state (PROCESSING, COMPLETED, FAILED).")
+    public ProcessingStatus getProcessingStatus(@PathParam("id") long id) {
```
willr3: This looks like it should be in a ProcessingService which has the single purpose of tracking the processing in h5m and not part of the ValueService that has the single purpose of enteracting with the values in h5m.

src/main/java/io/hyperfoil/tools/h5m/event/ChangeNotification.java
```diff
@@ -8,20 +10,24 @@
...
+ * @param nodeType      the detection node type
+ * @param changes       list of individual change results
  * @param configData    plugin-specific configuration (URL, email, channel, etc.), pre-parsed
  * @param configSecrets plugin-specific secrets (API tokens, passwords, etc.), pre-parsed
  * @param template      user-defined message template with placeholders, or null for default
  */
 public record ChangeNotification(
     String folderName,
+    long folderId,
+    long uploadId,
```
willr3: like Change I believe this should be a `valueId` because upload is a verb, not an entity.

src/main/java/io/hyperfoil/tools/h5m/notification/WebhookPlugin.java
```diff
@@ -125,9 +125,11 @@ public void validate(String configData) {
     private JqObject buildPayload(ChangeNotification notification) {
         JqObject.Builder payloadBuilder = JqObject.builder();
         payloadBuilder.put("folder", notification.folderName());
+        payloadBuilder.put("folderId", notification.folderId());
+        payloadBuilder.put("uploadId", notification.uploadId());
```
willr3: We should be consistent with `valueId` not `uploadId` because h5m does not have an upload entity

willr3 (commented):
two minor changes so that we use consistent language with users

PR #209, opened 2026-07-17
Add REST API key endpoints and bootstrap endpoint (issue #207)

stalep:
New REST endpoints for API key management:
  POST /api/apikey              — create key for authenticated user (returns raw key as text/plain)
  GET  /api/apikey              — list authenticated user's keys
  PUT  /api/apikey/{id}/revoke  — revoke a key (ownership check: must own key or be admin)

Bootstrap endpoint for first-time setup:
  POST /api/bootstrap           — creates admin user + API key when no users exist
                                  Returns BootstrapResponse with username and raw key
                                  Returns 409 if any user already exists

This enables external clients (Jenkins plugin, CI/CD tools) to:
1. Bootstrap h5m with security enabled (POST /api/bootstrap)
2. Use the returned API key for authenticated operations
3. Manage additional API keys via REST

New files:
- api/ApiKeyRequest.java — request DTO (description field)
- api/ApiKeyResponse.java — response DTO (id, description, dates, status)
- api/BootstrapResponse.java — bootstrap response (username, apiKey)
- rest/ApiKeyResource.java — API key CRUD endpoints
- rest/BootstrapResource.java — one-time bootstrap endpoint

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/rest/ApiKeyResource.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/rest/ApiKeyResource.java
@@ -1,0 +1,72 @@
+package io.hyperfoil.tools.h5m.rest;
+
+import io.hyperfoil.tools.h5m.api.ApiKey;
+import io.hyperfoil.tools.h5m.svc.ApiKeyService;
+import io.quarkus.security.Authenticated;
+import io.quarkus.security.identity.SecurityIdentity;
+import jakarta.inject.Inject;
+import jakarta.validation.constraints.NotEmpty;
+import jakarta.ws.rs.*;
+import jakarta.ws.rs.core.MediaType;
+import jakarta.ws.rs.core.Response;
+import org.eclipse.microprofile.openapi.annotations.Operation;
+import org.eclipse.microprofile.openapi.annotations.media.Content;
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+import org.eclipse.microprofile.openapi.annotations.parameters.Parameter;
+import org.eclipse.microprofile.openapi.annotations.responses.APIResponse;
+import org.eclipse.microprofile.openapi.annotations.tags.Tag;
+
+import java.util.List;
+
+@Path("/api/apikey")
+@Produces(MediaType.APPLICATION_JSON)
+@Tag(name = "API Key", description = "Manage API keys for authentication")
+public class ApiKeyResource {
+
+    @Inject
+    ApiKeyService apiKeyService;
+
+    @Inject
+    SecurityIdentity identity;
+
+    @POST
+    @Authenticated
+    @Operation(description = "Create a new API key for the authenticated user. " +
+            "The response includes the raw key in the 'rawKey' field — this is the only time it is displayed.")
+    @APIResponse(responseCode = "201", description = "API key created",
+            content = @Content(schema = @Schema(implementation = ApiKey.class)))
+    public Response create(
+            @QueryParam("description") @NotEmpty @Parameter(description = "Human-readable label for the key") String description) {
+        String username = identity.getPrincipal().getName();
+        ApiKeyService.CreateResult result = apiKeyService.createAndReturn(username, description);
+        return Response.status(Response.Status.CREATED).entity(result.apiKey()).build();
+    }
+
+    @GET
+    @Authenticated
+    @Operation(description = "List all API keys for the authenticated user.")
+    @APIResponse(responseCode = "200", description = "List of API keys",
+            content = @Content(schema = @Schema(implementation = ApiKey.class)))
+    public List<ApiKey> list() {
+        return apiKeyService.listByUser(identity.getPrincipal().getName());
+    }
+
+    @PUT
+    @Path("{id}/revoke")
+    @Authenticated
+    @Operation(description = "Revoke an API key. The key must belong to the authenticated user, or the user must be an admin.")
+    @APIResponse(responseCode = "204", description = "API key revoked")
+    @APIResponse(responseCode = "403", description = "Cannot revoke another user's key")
+    @APIResponse(responseCode = "404", description = "API key not found")
+    public void revoke(@PathParam("id") long keyId) {
+        ApiKey key = apiKeyService.getById(keyId);
+        if (key == null) {
+            throw new NotFoundException("API key not found: " + keyId);
+        }
+        String username = identity.getPrincipal().getName();
+        if (!key.owner().equals(username) && !identity.hasRole("admin")) {
+            throw new ForbiddenException("Cannot revoke another user's API key");
+        }
+        apiKeyService.revoke(keyId);
+    }
+}
--- a/src/main/java/io/hyperfoil/tools/h5m/rest/ApiKeyResource.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/rest/ApiKeyResource.java
@@ -8,7 +8,6 @@ import jakarta.inject.Inject;
 import jakarta.validation.constraints.NotEmpty;
 import jakarta.ws.rs.*;
 import jakarta.ws.rs.core.MediaType;
-import jakarta.ws.rs.core.Response;
 import org.eclipse.microprofile.openapi.annotations.Operation;
 import org.eclipse.microprofile.openapi.annotations.media.Content;
 import org.eclipse.microprofile.openapi.annotations.media.Schema;
@@ -33,13 +32,11 @@ public class ApiKeyResource {
     @Authenticated
     @Operation(description = "Create a new API key for the authenticated user. " +
             "The response includes the raw key in the 'rawKey' field — this is the only time it is displayed.")
-    @APIResponse(responseCode = "201", description = "API key created",
+    @APIResponse(responseCode = "200", description = "API key created",
             content = @Content(schema = @Schema(implementation = ApiKey.class)))
-    public Response create(
+    public ApiKey create(
             @QueryParam("description") @NotEmpty @Parameter(description = "Human-readable label for the key") String description) {
-        String username = identity.getPrincipal().getName();
-        ApiKeyService.CreateResult result = apiKeyService.createAndReturn(username, description);
-        return Response.status(Response.Status.CREATED).entity(result.apiKey()).build();
+        return apiKeyService.create(identity.getPrincipal().getName(), description);
     }
 
     @GET
--- a/src/main/java/io/hyperfoil/tools/h5m/svc/ApiKeyService.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/svc/ApiKeyService.java
@@ -41,22 +41,10 @@ public class ApiKeyService implements ApiKeyServiceInterface {
     @Inject
     ApiMapper apiMapper;
 
-    /** Result of creating an API key — holds the DTO (with rawKey populated) for the REST response. */
-    public record CreateResult(ApiKey apiKey, String rawKey) {}
-
     @Override
     @Transactional
-    public String create(String username, String description) {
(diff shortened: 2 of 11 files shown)


src/main/java/io/hyperfoil/tools/h5m/api/ApiKeyRequest.java
```diff
@@ -0,0 +1,8 @@
+package io.hyperfoil.tools.h5m.api;
+
+/**
+ * Request body for creating a new API key.
+ *
+ * @param description human-readable label for the key (e.g., "Jenkins CI", "local testing")
+ */
+public record ApiKeyRequest(String description) {}
```
barreiro: I acknowledge the use of a request object, but I don't think is needed in this case

src/main/java/io/hyperfoil/tools/h5m/api/ApiKeyResponse.java
```diff
@@ -0,0 +1,16 @@
+package io.hyperfoil.tools.h5m.api;
+
+import java.time.Instant;
+
+/**
+ * API key metadata returned when listing keys. Does not include the raw key
+ * (which is only returned once at creation time).
+ */
+public record ApiKeyResponse(
```
barreiro: should be named just `ApiKey`

src/main/java/io/hyperfoil/tools/h5m/api/BootstrapResponse.java
```diff
@@ -0,0 +1,7 @@
+package io.hyperfoil.tools.h5m.api;
+
+/**
+ * Response from the one-time bootstrap endpoint. Contains the admin username
+ * and the raw API key (displayed only once — cannot be retrieved again).
+ */
+public record BootstrapResponse(String username, String apiKey) {}
```
barreiro: this record is not needed

src/main/java/io/hyperfoil/tools/h5m/rest/BootstrapResource.java
```diff
@@ -0,0 +1,40 @@
...
+import jakarta.ws.rs.*;
+import jakarta.ws.rs.core.MediaType;
+import org.eclipse.microprofile.openapi.annotations.Operation;
+import org.eclipse.microprofile.openapi.annotations.tags.Tag;
+
+@Path("/api/bootstrap")
+@Produces(MediaType.APPLICATION_JSON)
+@Consumes(MediaType.APPLICATION_JSON)
+@Tag(name = "Bootstrap", description = "One-time system bootstrap for initial admin setup")
+public class BootstrapResource {
```
barreiro: I don't think this should be exposed in API. it's a detail.

we can bootstrap a bootstrap account in other ways. I suggest a startup event.

the logic would be similar. on startup it checks if there are no users create the bootstrap account and log a warning with the credential.

it should also check if default bootstrap username exists and if it does log a warning (mentioned that the bootstrap account should be removed).

src/main/java/io/hyperfoil/tools/h5m/rest/ApiKeyResource.java
```diff
@@ -0,0 +1,79 @@
...
+            throw new NotFoundException("API key not found: " + keyId);
+        }
+        // Ownership check: key must belong to caller, or caller must be admin
+        if (!key.user.username.equals(username) && !identity.hasRole("admin")) {
+            throw new ForbiddenException("Cannot revoke another user's API key");
+        }
+        apiKeyService.revoke(keyId);
+    }
+
+    private static ApiKeyResponse toResponse(ApiKey key, Instant now) {
```
barreiro: this should be a generated method in `ApiMapper` (and the existing `ApiKey` class should be named `ApiKeyEntity` , in line with all other entity classes)

src/main/java/io/hyperfoil/tools/h5m/rest/ApiKeyResource.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @Inject
+    SecurityIdentity identity;
+
+    @POST
+    @Authenticated
+    @Produces(MediaType.TEXT_PLAIN)
+    @Operation(description = "Create a new API key for the authenticated user. " +
+            "Returns the raw key as plain text — this is the only time it is displayed.")
+    public String create(ApiKeyRequest request) {
```
barreiro: I would rather see `ApikeyResponse` returned here, for consistency.

would need to be very careful to only include the key value in this endpoint and not in others. I don't think is hard though, as they key itself can be ignored at the mapper for the general case. in here we would either copy the key value over explicitly or have a different mapper that does not ignore the key.

src/main/java/io/hyperfoil/tools/h5m/rest/BootstrapResource.java
```diff
@@ -0,0 +1,40 @@
...
+
+    @POST
+    @PermitAll
+    @Operation(description = "One-time bootstrap: creates the initial admin user and API key. " +
+            "Only works when no users exist in the system. Returns the raw API key (displayed once, cannot be retrieved again).")
+    public BootstrapResponse bootstrap() {
+        if (User.count() > 0) {
+            throw new WebApplicationException("System already bootstrapped — users exist", 409);
+        }
+        String username = "admin";
```
barreiro: I'm a big fan of restricted namespaces !!

in this case I suggest `h5m.admin` to be the default bootstrap user name and also prevent creation of users with a name starting with `h5m.*`

barreiro (changes requested):
this can be simplified to just the essential (create / list / revoke). see my comments.

src/main/java/io/hyperfoil/tools/h5m/rest/BootstrapResource.java
stalep: Hm, how would that work against eg Jenkins client?

barreiro: the bootstrap is part of the setup. it allows the person installing h5m to create the first "real" user accounts.

it's not a concern of any client (like jenkins) to deal with this special account, and it should not be something exposed on the API

if you're thinking about scripting the install process, the bootstrap account password / key can be defined as an env property instead of random.

is there any use case I'm missing ??

stalep: Hm, env property should work yes

src/main/java/io/hyperfoil/tools/h5m/svc/ApiKeyService.java
```diff
@@ -82,4 +120,17 @@ static String hashKey(String rawKey) {
             throw new RuntimeException(e);
         }
     }
+
+    private static ApiKey toDto(ApiKeyEntity entity, Instant now, String rawKey) {
```
barreiro: this should be in `entity.mapper.ApiMapper` as one or possibly two methods (one with the raw key and another without)

src/main/java/io/hyperfoil/tools/h5m/svc/BootstrapService.java
```diff
@@ -0,0 +1,57 @@
...
+ * configured value.</p>
+ *
+ * <p>This enables automated environments (Testcontainers, CI/CD) to start
+ * h5m with security enabled and authenticate using a known API key without
+ * requiring OIDC or CLI-based user creation.</p>
+ */
+@ApplicationScoped
+public class BootstrapService {
+
+    private static final String BOOTSTRAP_USERNAME = "admin";
```
barreiro:
```suggestion
    private static final String BOOTSTRAP_USERNAME = "h5m.admin";
```

src/main/java/io/hyperfoil/tools/h5m/api/ApiKeyRequest.java
barreiro: I still think that this class is not needed. Also it would be interesting to enforce a `@NotEmpty` description in an attempt to give all keys in the system some meaning.

src/main/java/io/hyperfoil/tools/h5m/svc/BootstrapService.java
```diff
@@ -0,0 +1,57 @@
...
+ * {@code H5M_BOOTSTRAP_API_KEY} environment variable) is set and no users
+ * exist in the system, creates an admin user with an API key matching the
+ * configured value.</p>
+ *
+ * <p>This enables automated environments (Testcontainers, CI/CD) to start
+ * h5m with security enabled and authenticate using a known API key without
+ * requiring OIDC or CLI-based user creation.</p>
+ */
+@ApplicationScoped
+public class BootstrapService {
```
barreiro: A matter of personal preference, but this could be part of `ApiKeyService`

src/main/java/io/hyperfoil/tools/h5m/api/ApiKeyRequest.java
stalep: I added @NotEmpty, but kept ApiKeyRequest. I believe we'll extend it with expiry date etc later on.

barreiro: I'm going to insist on dropping `ApiKeyRequest`. it has no purpose, it adds nothing to the API

 the openAPI generators create these type of Requests objects regardless. it's just as extensible to edit the method signature as to edit such an object

I'd use `@QueryParam`  as it's consistent with the API already in place and it's easier to use (it's simpler to call `POST /api/apikey?description=Jenkins+CI` than to create an object)

src/main/java/io/hyperfoil/tools/h5m/rest/ApiKeyResource.java
```diff
@@ -0,0 +1,72 @@
...
+    @Authenticated
+    @Operation(description = "Create a new API key for the authenticated user. " +
+            "The response includes the raw key in the 'rawKey' field — this is the only time it is displayed.")
+    @APIResponse(responseCode = "201", description = "API key created",
+            content = @Content(schema = @Schema(implementation = ApiKey.class)))
+    public Response create(
+            @QueryParam("description") @NotEmpty @Parameter(description = "Human-readable label for the key") String description) {
+        String username = identity.getPrincipal().getName();
+        ApiKeyService.CreateResult result = apiKeyService.createAndReturn(username, description);
+        return Response.status(Response.Status.CREATED).entity(result.apiKey()).build();
```
barreiro:
```suggestion
    @APIResponse(responseCode = "200", description = "API key created",
            content = @Content(schema = @Schema(implementation = ApiKey.class)))
    public ApiKey create(
            @QueryParam("description") @NotEmpty @Parameter(description = "Human-readable label for the key") String description) {
        return apiKeyService.create(identity.getPrincipal().getName(), description);
```
although technically the POST is supposed to return 201, in practice it's not checked at the client and many APIs return 200 as the status code for simplicity.

other POST endpoints on h5m already follow that pattern.

If we decide to follow the spec, then returning `RestResponse<ApiKey>` is the quarkus way of ensuring type safety on the response.

src/main/java/io/hyperfoil/tools/h5m/svc/ApiKeyService.java
```diff
@@ -19,41 +27,84 @@
...
+    Optional<String> bootstrapApiKey;
+
     @Inject
     UserService userService;
 
+    @Inject
+    ApiMapper apiMapper;
+
+    /** Result of creating an API key — holds the DTO (with rawKey populated) for the REST response. */
+    public record CreateResult(ApiKey apiKey, String rawKey) {}
```
barreiro:
```suggestion
```
we just need to have 2 `create` methods: one internal where the key is specified; another that generates the key and calls the first one.

both should return `ApiKey`.

barreiro (approved):

PR #210, opened 2026-07-20
fix Work recalculate concurrency with cascade and fix rd work

willr3:
This introduces Work.cascade which prevents work from creating more work when it creates new Values. This option is used to recalculate ephemeral Values for recalculate without incurring the performance penalty of wasted additional computation for unnecessary nodes and sorting. 
This also fixes recalculate to work concurrently with upload and other recalculates that could impact the same scope. The RecalculateWorkQueueTest uses a WorkExecutor without threads to manually verify the queuing.
This PR also removes recalculate for a full folder.

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/api/NodeType.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/api/NodeType.java
@@ -2,29 +2,34 @@ package io.hyperfoil.tools.h5m.api;
 
 import org.eclipse.microprofile.openapi.annotations.media.Schema;
 
+import java.util.Arrays;
+import java.util.stream.Collectors;
+
 @Schema(description = "The type of transformation a node performs")
 public enum NodeType {
-    EDIVISIVE("ed"),
-    FINGERPRINT("fp"),
-    FIXED_THRESHOLD("ft"),
-    JQ("jq"),
-    JS("js"),
-    JSONATA("nata"),
-    RELATIVE_DIFFERENCE("rd"),
-    ROOT("root"),
-    STDDEV_ANOMALY("sd"),
-    SPLIT("split"),
-    USER_INPUT("user");
+    EDIVISIVE("ed",true),
+    FINGERPRINT("fp",true),
+    FIXED_THRESHOLD("ft",true),
+    JQ("jq",false),
+    JS("js",false),
+    JSONATA("nata",false),
+    RELATIVE_DIFFERENCE("rd",true),
+    ROOT("root",false),
+    STDDEV_ANOMALY("sd",true),
+    SPLIT("split",false),
+    USER_INPUT("user",false);
 
     private final String display;
+    private final boolean analysis;
 
-    NodeType(String display) {
-        this.display = display;
+    NodeType(String display,boolean analysis) {
+        this.display = display;this.analysis = analysis;
     }
 
     public String display() {
         return display;
     }
+    public boolean isAnalysis(){return analysis;}
 
     /**
      * Returns true if this node type is a change detection node.
--- a/src/main/java/io/hyperfoil/tools/h5m/entity/work/Work.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/entity/work/Work.java
@@ -15,19 +15,29 @@ import java.util.stream.Collectors;
 //custom post nodegroup actions could have sourceNodes without activeNode
 public class Work implements Runnable, Comparable<Work>{
 
-    public List<Long> sourceValueIds;//IDs of source values — full entities are loaded in WorkService.execute()
+    private List<Long> sourceValueIds;//IDs of source values — full entities are loaded in WorkService.execute()
 
-    public List<NodeEntity> sourceNodes; //what is going to use a list of sources that are not already listed for the activeNode?
+    private List<NodeEntity> sourceNodes; //what is going to use a list of sources that are not already listed for the activeNode?
 
-    public int retryCount;
+    private int retryCount;
 
-    public Set<NodeEntity> activeNodes;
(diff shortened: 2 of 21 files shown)


src/main/java/io/hyperfoil/tools/h5m/api/NodeType.java
```diff
@@ -2,29 +2,34 @@
...
-    JQ("jq"),
-    JS("js"),
-    JSONATA("nata"),
-    RELATIVE_DIFFERENCE("rd"),
-    ROOT("root"),
-    STDDEV_ANOMALY("sd"),
-    SPLIT("split"),
-    USER_INPUT("user");
+    EDIVISIVE("ed",true),
+    FINGERPRINT("fp",true),
```
Arjun-Parmani, line 11: Why finger print node is marked as analysis node along with detection node? Asking since it is used for aggregating. Is it because detection nodes need them for grouping?

src/main/java/io/hyperfoil/tools/h5m/entity/work/Work.java
Arjun-Parmani: far as I've checked setActiveNodes() is not called but should this also requires a clean up here?

src/test/java/io/hyperfoil/tools/h5m/queue/WorkQueueTest.java
```diff
@@ -29,6 +30,42 @@ public class WorkQueueTest extends FreshDb {
...
+
+        Work one = new Work(node,node.sources,Collections.emptyList());
+        System.out.println("one="+one.hashCode());
+        q.add(one);
+
+        tm.begin();
+        NodeEntity found = NodeEntity.findById(node.id);
+        found.operation = ".node + 1";
+        found.persist();
+        System.out.println("found="+found);
```
Arjun-Parmani: Should there be proper asserts for the test?

src/test/java/io/hyperfoil/tools/h5m/cli/H5mTest.java
```diff
@@ -597,42 +597,6 @@ public void upload_js_multi_input(QuarkusMainLauncher launcher) throws IOExcepti
         assertFalse(result.getOutput().contains("null")||result.getOutput().contains("NULL"),"list values should not contain null\n"+result.getOutput());
     }
 
-    @Test
```
Arjun-Parmani, line 600: Do we have a CLI test for per node re calculation?

willr3: No, this PR was to correct the functionality in how work is handled. Once this is merged we can explore if it needs to be exposed to users via a CLI or if it should trigger automatically when users perform related actions (e.g. adding or updating a node).

src/test/java/io/hyperfoil/tools/h5m/queue/WorkQueueTest.java
willr3: The `omg` test was something I created to test while making code changes but was not meant to be included in the PR. I removed it

src/main/java/io/hyperfoil/tools/h5m/api/NodeType.java
willr3: An analysis node is one that analyzes the output of other nodes and does not make a node ephemeral.

src/main/java/io/hyperfoil/tools/h5m/entity/work/Work.java
willr3: >I've checked setActiveNodes() is not called

Two things happened here. `Work` was previously a Panache Entity. Panache allows developers to write code that references public member variables and it will replace those references with getters and setters at compile time. This should have been fixed when Work was changed to a plain java object (POJO) but good catch that it wasn't performed. I fixed `Work` to no longer have public member variables and updated the PR.

>also requires a clean up here?

When you say "clean up" it leaves me unsure of what you mean. I assume you mean that `RelativeDifference` should be removed from the check to make the Work cumulative and if so you're correct. I updated the method.

src/main/java/io/hyperfoil/tools/h5m/svc/WorkService.java
```diff
@@ -36,6 +36,14 @@
 @ApplicationScoped
 public class WorkService implements WorkServiceInterface {
 
+    static void main() {
```
stalep: Is this main() block used anywhere?

src/main/java/io/hyperfoil/tools/h5m/svc/ProcessingService.java
```diff
@@ -68,7 +71,7 @@ public void recoverIncompleteProcessing(@Observes @Priority(2) StartupEvent ev)
                 for (ProcessingTrackerEntity tracking : incomplete) {
                     switch (tracking.type) {
                         case UPLOAD -> recoverUpload(tracking);
-                        case RECALCULATE -> recoverRecalculate(tracking);
+                        //case RECALCULATE -> recoverRecalculate(tracking);
```
stalep: remove?

stalep (approved):

src/main/java/io/hyperfoil/tools/h5m/svc/WorkService.java
willr3: :laughing:  that is from when I was testing something and forgot to delete it. I will clean it up

src/main/java/io/hyperfoil/tools/h5m/svc/ProcessingService.java
willr3: fixed

PR #211, opened 2026-07-21
Fixed folder delete when a folder has an upload

Arjun-Parmani:
Deleting a folder that has uploads was failing with a 500 due to a FK constraint violation on the value table. The bulk delete bypasses JPA cascade, so dependent rows were not cleaned up before the folder row was deleted.

  Changes:
  - FolderService.delete() now explicitly cleans up all dependent rows in the correct FK order before deleting the folder
  - Added delete_folder_with_uploads test in FolderServiceTest to cover this scenario 

Close issue #204


willr3 (changes requested):
There are two flaws with this:
1. The implemented method is an abstraction violation and we should not encourage that in our software.
2. This is now what we discussed for how to handle folder delete. The Folder.delete should not remove the underlying group so that delete is a two stage operation.

PR #217, opened 2026-07-24
Fix stale ancestorCache in NodeEntity.dependsOn (issue #215)

stalep:
Move ancestor cache from NodeEntity to Work. NodeEntity.dependsOn() now always traverses the source graph, avoiding stale results when a parent node's sources are modified. Work.precomputeAncestors() caches ancestor IDs on the short-lived Work object for use in afterCompletion where lazy source traversal would fail.

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/entity/NodeEntity.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/entity/NodeEntity.java
@@ -65,8 +65,7 @@ public abstract class NodeEntity extends PanacheEntity implements Comparable<Nod
 
     public List<NodeEntity> getSources() {return sources;}
 
-    @Transient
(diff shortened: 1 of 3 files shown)


willr3 (changes requested):
1. This removes the race condition on `NodeEntity` but I believe it creates the same race condition for `Work`. The commit mentions "short lived" as a justification for the design but I do not believe we should introduce race conditions in a core part of the computation logic. 
2. This commit takes authorship credit for the work done by another engineer (myself). That discourages commitment from engineers who invest the time to investigate a bug, identify the source, and create a unit test that succinctly reproduces the issue. If we want people to contribute we should not take credit for their work. What method of contribution would you prefer when engineers invest the time to identify bugs and create tests for existing code in the repository?

PR #219, opened 2026-07-27
add ancestor value filtering to valueService.getGroupedValues

willr3:


src/test/java/io/hyperfoil/tools/h5m/FreshDb.java
```diff
@@ -25,7 +25,7 @@ public class FreshDb {
     String dbKind;
 
     @BeforeEach
-    @AfterEach
+    //@AfterEach
```
stalep: Is this intentional?

src/main/java/io/hyperfoil/tools/h5m/svc/ValueService.java
```diff
@@ -512,52 +537,53 @@ root_sort as (
         }
 
         String filter = !nodeFilter.isEmpty() || !fingerPrintWhere.isEmpty() ? "where "+ nodeFilter + (!nodeFilter.isEmpty() && !fingerPrintWhere.isEmpty() ? " and " : "" ) + fingerPrintWhere : "";
+        String queryStr = (switch (dbKind) {
```
stalep: the two sql's looks identical, can we remove the case check?

src/main/java/io/hyperfoil/tools/h5m/svc/ValueService.java
```diff
@@ -469,15 +469,39 @@ public List<JqValue> getGroupedValues(Long nodeId){
 
     @Override
     @Transactional
-    public List<JqValue> getGroupedValues(Long nodeId, List<Long> filterNodeIds, Map<Long,JqValue> fingerprints, Long sortByNodeId) {
+    public List<JqValue> getGroupedValues(Long nodeId, Long valueId, List<Long> filterNodeIds, Map<Long,JqValue> fingerprints, Long sortByNodeId) {
```
stalep, line 469: Should we keep the old signature as well which would call the new with valueId null?

src/main/java/io/hyperfoil/tools/h5m/svc/ValueService.java
willr3: They are not identical, the json related functions are different on sqlite and postgresql

src/test/java/io/hyperfoil/tools/h5m/FreshDb.java
willr3: good catch! I will fix

src/main/java/io/hyperfoil/tools/h5m/svc/ValueService.java
willr3: sure, I will add it back to ValueService and ValueServiceInterface

PR #220, opened 2026-07-29
Rename User entity to UserEntity and introduce User DTO

stalep:
Rename entity/User to entity/UserEntity for consistency with the project naming convention (ApiKeyEntity, FolderEntity, ValueEntity).

Move Role enum from entity to api package. Create User record DTO in the api package. UserServiceInterface now returns User DTOs instead of JPA entities, keeping the persistence layer encapsulated.

UserService maps entities to DTOs internally via toDto(). All consumers (CLI commands, security providers, tests) use the User DTO with record accessors. No database schema changes — @Entity(name = "h5m_user") is unchanged.

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/svc/UserService.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/svc/UserService.java
@@ -1,9 +1,12 @@
 package io.hyperfoil.tools.h5m.svc;
 
+import io.hyperfoil.tools.h5m.api.Role;
+import io.hyperfoil.tools.h5m.api.User;
 import io.hyperfoil.tools.h5m.api.svc.UserServiceInterface;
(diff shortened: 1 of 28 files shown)


src/main/java/io/hyperfoil/tools/h5m/svc/UserService.java
```diff
@@ -43,6 +46,10 @@ public void setRole(long userId, Role role) {
     @Override
     @Transactional
     public long count() {
-        return User.count();
+        return UserEntity.count();
+    }
+
+    private static User toDto(UserEntity entity) {
```
barreiro: this should be Mapstruct generated in `entity.mapper.ApiMapper`

barreiro (commented):
looks good. should probably be extended to the `Team` entity >> `TeamEntity` as well !!

barreiro (approved):

PR #234, opened 2026-08-04
Upload data UI

Arjun-Parmani:
**Problem statement**
The h5m web UI lacked the ability for users to upload JSON data to folders directly from the browser. Added Upload data functionality in the UI.

**Description**
Added a new upload modal to the folder view supporting multiple input modes -  file upload, raw JSON paste, and remote URL. Integrated the application-wide toast notification system for consistent user feedback across all upload outcomes and corrected failing REST endpoint tests to align with the current API contract.

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/rest/FolderResource.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/rest/FolderResource.java
@@ -1,6 +1,7 @@
 package io.hyperfoil.tools.h5m.rest;
 
 import io.hyperfoil.tools.jjq.value.JqValue;
+import io.hyperfoil.tools.jjq.value.JqValues;
 import io.hyperfoil.tools.h5m.api.Folder;
 import io.hyperfoil.tools.h5m.api.FolderSummary;
 import io.hyperfoil.tools.h5m.api.RecalculationStatus;
@@ -8,23 +9,40 @@ import io.hyperfoil.tools.h5m.api.svc.FolderServiceInterface;
 import io.hyperfoil.tools.h5m.api.svc.ValueServiceInterface;
 import io.hyperfoil.tools.h5m.svc.RecalculationService;
 import io.hyperfoil.tools.h5m.svc.RecalculationTracker;
+import io.quarkus.runtime.configuration.MemorySize;
 import io.quarkus.security.Authenticated;
 import jakarta.annotation.security.PermitAll;
 import jakarta.inject.Inject;
 import jakarta.validation.constraints.NotNull;
 import jakarta.ws.rs.*;
 import jakarta.ws.rs.core.MediaType;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
 import org.eclipse.microprofile.openapi.annotations.Operation;
-import org.eclipse.microprofile.openapi.annotations.parameters.Parameter;
+import org.eclipse.microprofile.openapi.annotations.responses.APIResponse;
 import org.eclipse.microprofile.openapi.annotations.tags.Tag;
+import org.jboss.resteasy.reactive.RestForm;
+import org.jboss.resteasy.reactive.multipart.FileUpload;
 
+import java.io.IOException;
+import java.net.URL;
+import java.net.URLConnection;
+import java.nio.charset.StandardCharsets;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Stream;
+
+import static jakarta.ws.rs.core.MediaType.MULTIPART_FORM_DATA;
+import static java.nio.file.Files.readAllBytes;
 
 @Path("/api/folder")
 @Produces(MediaType.APPLICATION_JSON)
 @Tag(name = "Folder", description = "Manage folders for uploaded data")
 public class FolderResource {
+
+    @ConfigProperty(name = "quarkus.http.limits.max-body-size")
+    MemorySize maxBodySize;
 
     @Inject
     FolderServiceInterface folderService;
@@ -83,15 +101,54 @@ public class FolderResource {
 
     @POST
     @Path("{id}/upload")
+    @Consumes(MULTIPART_FORM_DATA)
     @Authenticated
     @Operation(description = "Upload JSON data to a folder. Returns immediately with an uploadId.")
+    @APIResponse(responseCode = "200", description = "Upload successful, returns uploadId")
+    @APIResponse(responseCode = "400", description = "Request received but content is not valid JSON or URL scheme is not http/https")
     public long upload(
             @PathParam("id") long id,
-            JqValue data) {
-        if (data == null) {
-            throw new BadRequestException("Missing request body");
+            @RestForm("raw") String raw,
+            @RestForm("url") URL url,
+            @RestForm("file") FileUpload file) {
+
+        if (Stream.of(url, raw == null || raw.isBlank() ? null : raw, file).filter(Objects::nonNull).count() != 1) {
+            throw new BadRequestException("Provide exactly one of 'file', 'raw', or 'url'");
         }
-        return folderService.upload(id, data).uploadId;
+
+        byte[] bytes;
+
+        try {
+            if (url != null) {
+                if (!Set.of("http", "https").contains(url.getProtocol())) {
+                    throw new BadRequestException("Only http/https URLs are allowed");
+                }
+                URLConnection connection = url.openConnection();
+                connection.setConnectTimeout(5000);
+                connection.setReadTimeout(30000);
+                int readLimit = (int) Math.min(maxBodySize.asLongValue() + 1, Integer.MAX_VALUE);
+                try (var inputStream = connection.getInputStream()) {
+                    bytes = inputStream.readNBytes(readLimit);
+                }
+                if (bytes.length > maxBodySize.asLongValue()) {
+                    throw new BadRequestException("Content at '" + url + "' exceeds the maximum upload size");
+                }
+            } else if (file != null) {
+                bytes = readAllBytes(file.uploadedFile());
+            } else {
+                bytes = raw.getBytes(StandardCharsets.UTF_8);
+            }
+        } catch (BadRequestException e) {
+            throw e;
+        } catch (IOException e) {
+            throw new BadRequestException("Failed to read upload data: " + e.getMessage());
+        }
+
+        try {
+            return folderService.upload(id, JqValues.parse(bytes)).uploadId;
+        } catch (Exception e) {
+            throw new BadRequestException("Invalid JSON: " + e.getMessage());
+        }
     }
 
     @GET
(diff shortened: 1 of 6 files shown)


src/main/java/io/hyperfoil/tools/h5m/api/UploadRequest.java
```diff
@@ -0,0 +1,10 @@
+package io.hyperfoil.tools.h5m.api;
+
+import io.hyperfoil.tools.jjq.value.JqValue;
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+
+@Schema(description = "Request body for uploading JSON data to a folder")
+public record UploadRequest(
```
barreiro: this object is not needed. I would rather see the different types uploads to be sent as `MULTIPART_FORM_DATA` (openAPI generators will create some kind of request object from it)

this would be what the endpoint would look like: 
```
import static jakarta.ws.rs.core.MediaType.MULTIPART_FORM_DATA;

import org.jboss.resteasy.reactive.RestForm;
import org.jboss.resteasy.reactive.multipart.FileUpload;

    @POST
    @Path("{id}/upload")
    @Consumes(MULTIPART_FORM_DATA)
    @Operation(description = "Upload JSON data to a folder. Returns immediately with an uploadId.")
    @APIResponse(responseCode = "400", description = "Request received but context is not valid format")
    @Authenticated
    long upload(@RestForm("raw") String raw, @RestForm("url") URL url, @RestForm("file") FileUpload file) {
        if (raw != null) { ... }
        if (url != null) { ... }
        if (file != null) { ... }
}
```

Tests will require `io.restassured` multipart builders instead of simple `.body(json)`.

src/main/java/io/hyperfoil/tools/h5m/rest/FolderResource.java
```diff
@@ -87,9 +96,29 @@ public void deleteFolder(@PathParam("id") long id) {
     @Operation(description = "Upload JSON data to a folder. Returns immediately with an uploadId.")
     public long upload(
             @PathParam("id") long id,
-            JqValue data) {
+            UploadRequest body) {
+        String url = body != null ? body.url() : null;
+        JqValue data = null;
+        if (url != null && !url.isEmpty()) {
+            try (var inputStream = URI.create(url).toURL().openStream()) {
```
barreiro: for the URL parameter, verify that it's schema is HTTP / HTTPS to prevent some SSRF attacks that look into the local filesystem (like file://, jar://) and other protocols like ftp://
```
  if (!Set.of("http", "https").contains(uri.getScheme())) {                                                                                                                        
      throw new BadRequestException("Only http/https URLs are allowed");                                                                                                           
  } 
```

src/main/webui/src/app/components/UploadDataModal.tsx
```diff
@@ -0,0 +1,404 @@
...
+                      {selectedFiles.length} files selected
+                    </p>
+                  )}
+                </>
+              )}
+            </TabPanel>
+
+            <TabPanel>
+              <TextInput
+                id="paste-name"
```
barreiro, line 266: I don't think requiring a name for a JSON upload makes sense. it's not something h5m has in it's model and from what I can understand is only used for the user of the Web UI track it's uploads. 

Should just give it a generic one, like `JSON` or `JSON #1`.

src/main/webui/src/app/components/DataTab.tsx
```diff
@@ -91,6 +97,8 @@ export const DataTab = ({ folderId, groupId }: { folderId: number; groupId: numb
   const [configModalOpen, setConfigModalOpen] = useState(false);
   const [editingView, setEditingView] = useState<View | null>(null);
   const [modalKey, setModalKey] = useState(0);
+  const [uploadModalOpen, setUploadModalOpen] = useState(false);
+  const [recentUploads, setRecentUploads] = useState<Array<{ fileName: string; uploadId: number }>>([]);
```
barreiro: could also track the upload time

src/main/webui/src/app/components/UploadDataModal.tsx
```diff
@@ -0,0 +1,404 @@
...
+    setCurrentStatus(null);
+    setIsDragging(false);
+    setPasteName('');
+    setPasteText('');
+    setUrlInput('');
+    upload.reset();
+    onClose();
+  }
+
+  function extractMsg(e: AxiosError<Error>): string {
```
barreiro: this is handle on the notification system by using the `handleError(msg, e);` method.

src/main/webui/src/app/components/UploadDataModal.tsx
```diff
@@ -0,0 +1,404 @@
...
+      if (successCount > 0) {
+        notifications.success(`${successCount} file${successCount > 1 ? 's' : ''} uploaded successfully`);
+      }
+
+    } else if (activeTab === TAB_PASTE) {
+      let parsed: unknown;
+      try {
+        parsed = JSON.parse(pasteText);
+      } catch {
+        notifications.error('Invalid JSON', 'Please check the pasted content.');
```
barreiro: This should change the `invalidText` of the `TextInput`

The notifications are to display info to the use after the modal is closed.

src/main/webui/src/app/components/UploadDataModal.tsx
```diff
@@ -0,0 +1,404 @@
...
+          <TabList aria-label="Upload mode">
+            <Tab>File Upload</Tab>
+            <Tab>Paste JSON</Tab>
+            <Tab>URL</Tab>
+          </TabList>
+          <TabPanels>
+
+            <TabPanel>
+              {currentIndex !== null ? (
+                <div style={{
```
barreiro: This tab should be using a carbon `FileUploader` component instead of trying to write one !!

src/main/webui/src/app/components/UploadDataModal.tsx
barreiro: other option is to close the modal and display the notification. it will depend when you perform the validation (on change or on upload) and when when you want to give feedback to the user.

src/main/webui/src/app/components/UploadDataModal.tsx
```diff
@@ -0,0 +1,404 @@
...
+    upload.reset();
+    onClose();
+  }
+
+  function extractMsg(e: AxiosError<Error>): string {
+    const msg = e.response?.data as string | undefined;
+    return typeof msg === 'string' && msg.length > 0 ? msg : (e.message ?? 'Upload failed');
+  }
+
+  async function handleUpload() {
```
barreiro, line 83: this method should just go over `selectedFiles`, `pastText` and `urlInput` and update all those one by one.

for each create a promise to handle the result of the mutate call. the point is that all should be handled the same way.

barreiro (changes requested):
Overall is good, but I see a few points that can be simplified / optimized / improved.

src/main/webui/src/app/components/UploadDataModal.tsx
Arjun-Parmani: The main idea was that adding a dedicated text field enables users to apply meaningful labels to pasted JSON payloads rather than relying on generic sequential identifiers.

src/main/webui/src/app/components/UploadDataModal.tsx
Arjun-Parmani: Makes sense.  I'll use the invalid / invalidText prop on the TextArea to show the error inline on the field itself, and remove the notification. I think closing a Modal and then throwing a toast notification just adds extra steps for users to again paste the JSON payload.

src/main/java/io/hyperfoil/tools/h5m/rest/FolderResource.java
```diff
@@ -83,13 +93,52 @@ public void deleteFolder(@PathParam("id") long id) {
...
+            @RestForm("url") URL url,
+            @RestForm("file") FileUpload file) {
+
+        JqValue data = null;
+
+        if (url != null) {
+            if (!Set.of("http", "https").contains(url.getProtocol())) {
+                throw new BadRequestException("Only http/https URLs are allowed");
+            }
+            try (var inputStream = url.openStream()) {
```
stalep: Hm, what if this file is huge?
It might not be a real world issue, but would also be good to check for?

src/main/java/io/hyperfoil/tools/h5m/rest/FolderResource.java
```diff
@@ -83,13 +93,52 @@ public void deleteFolder(@PathParam("id") long id) {
...
+                String content = new String(inputStream.readAllBytes(), StandardCharsets.UTF_8);
+                data = JqValues.parse(content);
+            } catch (IOException e) {
```
stalep, line 128: Hm, there is a fairly long chain of catch re-throw's here, looks ok, but just not "pretty" :)

stalep (commented):
Looks ok, some minor comments only.

src/main/java/io/hyperfoil/tools/h5m/rest/FolderResource.java
Arjun-Parmani: Good observation! readAllBytes() on the URL stream has no size cap. May be adding a limit here for reading the bytes would be a good guard? Anything exceeding the guard/boundary can be errored out.

stalep: Hm, I guess there are several methods of doing it, one could be to try to get the HEAD req first to get the Content-Lenght without downloading the body;

```
HttpURLConnection conn = (HttpURLConnection) url.openConnection();
conn.setRequestMethod("HEAD");
conn.setConnectTimeout(5000);
conn.setReadTimeout(5000);
long contentLength = conn.getContentLengthLong();
if (contentLength > MAX_UPLOAD_SIZE) {
    throw new BadRequestException("Remote resource too large: " + contentLength + " bytes (max " + MAX_UPLOAD_SIZE + ")");
}
```

Not all servers return Content-Length, so in any case I think would be better than url.OpenStream() is to use the jvm builtin HttpClient. It got a lot better the last few years. We could also use the vert.x httpClient, but imo the builtin client is just fine;

```
HttpClient client = HttpClient.newBuilder()
    .connectTimeout(Duration.ofSeconds(5))
    .followRedirects(HttpClient.Redirect.NORMAL)
    .build();

HttpRequest request = HttpRequest.newBuilder()
    .uri(url.toURI())
    .timeout(Duration.ofSeconds(30))
    .GET()
    .build();

HttpResponse<String> response = client.send(request, HttpResponse.BodyHandlers.ofString());

if (response.statusCode() != 200) {
    throw new BadRequestException("URL returned HTTP " + response.statusCode());
}

data = JqValues.parse(response.body());
```

You could also stream instead of reading everything into a string by;

```
HttpResponse<InputStream> response = client.send(request, HttpResponse.BodyHandlers.ofInputStream());

long maxSize = 100 * 1024 * 1024; // 100MB limit
byte[] bytes;
try (InputStream is = response.body()) {
    bytes = is.readNBytes((int) maxSize + 1);
    if (bytes.length > maxSize) {
        throw new BadRequestException("Remote resource exceeds " + maxSize + " bytes");
    }
}
data = JqValues.parse(bytes);
```

This limits to 100mb, it could probably be a bit larger for h5m I guess?

barreiro (changes requested):
Another round of comments !!

barreiro (changes requested):

barreiro (approved):

PR #235, opened 2026-08-04
fix jq path to match postgresql lax jsonpath for load-legacy-tests

willr3:
This fixes issue #230 by changing how jq operations are created for Horreum extractor paths. The new jq operations will check each key in a chain of keys to handle the auto array expansion that occurs in postgresql lax jsonpath.

### DIFF
--- a/src/test/java/io/hyperfoil/tools/h5m/cli/LoadLegacyTestsTest.java
+++ b/src/test/java/io/hyperfoil/tools/h5m/cli/LoadLegacyTestsTest.java
@@ -1,6 +1,14 @@
 package io.hyperfoil.tools.h5m.cli;
 
+import io.hyperfoil.tools.h5m.FreshDb;
+import io.hyperfoil.tools.h5m.api.Folder;
+import io.hyperfoil.tools.h5m.api.Upload;
+import io.hyperfoil.tools.h5m.entity.ValueEntity;
+import io.hyperfoil.tools.h5m.svc.ValueService;
+import io.hyperfoil.tools.jjq.JqProgram;
 import io.hyperfoil.tools.jjq.value.JqObject;
+import io.hyperfoil.tools.jjq.value.JqString;
+import io.hyperfoil.tools.jjq.value.JqValue;
 import io.hyperfoil.tools.jjq.value.JqValues;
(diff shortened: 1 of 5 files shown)


src/test/java/io/hyperfoil/tools/h5m/cli/LoadLegacyTestsTest.java
```diff
@@ -14,23 +22,119 @@
...
+    FolderService folderService;
+
+    @Inject
+    ValueService valueService;
+
+    @Inject
+    TransactionManager tm;
+
+    @Test
+    public void workers(){
```
stalep: This could be removed I think.

src/test/java/io/hyperfoil/tools/h5m/cli/LoadLegacyTestsTest.java
```diff
@@ -345,7 +449,7 @@ public void createNodesFromLabel_multi_extractor_single_param_direct_sources(){
         LoadLegacyTests.Extractor workload = new LoadLegacyTests.Extractor("workload","$.workload",false);
         LoadLegacyTests.Extractor results = new LoadLegacyTests.Extractor("results","$.results.*",true);
         LoadLegacyTests.Label label = new LoadLegacyTests.Label(-1,"Autobench",
-                "v => v[\"results\"].reduce((a,b) => a+b) / v[\"results\"].length",
+                "v => v[\"results\"].reduce((a,b) => a+b) / v[\"results\"].stop",
```
stalep: why was this changed?

stalep (commented):
Looks good, I just had two minor comments

src/test/java/io/hyperfoil/tools/h5m/cli/LoadLegacyTestsTest.java
willr3: because I failed with the scope of a variable rename and thought I fixed them all but missed one :)

src/test/java/io/hyperfoil/tools/h5m/cli/LoadLegacyTestsTest.java
willr3: removed along with the unused valueService

PR #241, opened 2026-08-06
Add centralized notification infrastructure for the UI

barreiro:
##  Summary

- Add a `NotificationContext` / `NotificationProvider` / `useNotification` hook that provides a centralized way to show toast notifications across the app, using Carbon's `ToastNotification` component
- The provider supports info, success, warning, and error shorthands, a generic handleError that extracts messages from Axios-like errors without coupling to Axios, and update for in-place modification (e.g. progress bars)
- Includes deduplication, a max-10 cap to prevent screen flooding, and sticky errors (no auto-dismiss) while info/success/warning auto-dismiss after 10 seconds
- Migrate the delete folder flow in `DashboardPage` as a first consumer: replaces inline error state with toast notifications on success and failure

## Sample

Screenshot_2026-08-06_04-26-49

### DIFF
--- a/src/main/webui/src/app/context/NotificationProvider.tsx
+++ b/src/main/webui/src/app/context/NotificationProvider.tsx
@@ -1,0 +1,121 @@
+import { DEFAULT_NOTIFICATION_TIMEOUT, Notification, NotificationContext, NotificationKind, NotificationUpdate } from '@app/context/NotificationContext.tsx';
+import { ToastNotification } from '@carbon/react';
+import { ReactNode, useCallback, useEffect, useMemo, useReducer, useRef } from 'react';
+
+const MAX_NOTIFICATIONS = 10;
+
+type Action = { type: 'add'; notification: Notification } | { type: 'update'; id: number; fields: NotificationUpdate } | { type: 'remove'; id: number };
+
+function reducer(state: Notification[], action: Action): Notification[] {
+  switch (action.type) {
+    case 'add': {
+      const next = [action.notification, ...state];
+      return next.length > MAX_NOTIFICATIONS ? next.slice(0, MAX_NOTIFICATIONS) : next;
+    }
+    case 'update':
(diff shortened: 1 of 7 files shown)


src/main/webui/src/app/context/NotificationProvider.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+const MAX_NOTIFICATIONS = 10;
+
+type Action = { type: 'add'; notification: Notification } | { type: 'update'; id: number; fields: NotificationUpdate } | { type: 'remove'; id: number };
+
+function reducer(state: Notification[], action: Action): Notification[] {
+  switch (action.type) {
+    case 'add': {
+      // dedup logic in the reducer that skips the add if a notification with the same kind + title + subtitle already exists in the array.
+      if (state.some((n) => n.kind === action.notification.kind && n.title === action.notification.title && n.subtitle === action.notification.subtitle)) {
```
Arjun-Parmani: I have a question here: 
When add() is called for a duplicate notification (matching kind, title, and subtitle), below in the fucntion add(...), it increments nextId.current and returns a newly generated ID (e.g., 2). However, inside the reducer, the duplicate check triggers return state; which discards the new notification object.
add() hands back 2 to the calling component, but the state array only contains the original notification (1). If the caller later attempts to dismiss or update the toast using that returned ID (notifications.remove(id) or notifications.update(id, ...)), the call silently fails because ID 2 does not exist in state leaving the original notification stuck on screen.

barreiro: Good catch. Moved the dedup check to `add()` instead of the reducer. 

It uses a ref to access current state, so when a duplicate is found it returns the existing notification's ID without incrementing the counter or dispatching. 
The returned ID is always valid for subsequent `update()` or `remove()` calls.

PR #245, opened 2026-08-06
add to /api/value/{id}/descendants filtering options

willr3:


src/main/java/io/hyperfoil/tools/h5m/svc/ValueService.java
```diff
@@ -688,6 +688,53 @@ WITH RECURSIVE sourceRecursive (v_id) AS (
...
 
+    /**
+     * returns the values that depend on the root value somewhere up the hierarchy and come from the specified node
+     * @param rootValueId
+     * @param nodeId
+     * @return
+     */
+    @Transactional
+    @SuppressWarnings("unchecked")
+    public List<Value> getDescendantValues(long rootValueId,long nodeId){
```
barreiro: this method can be removed and `getDescendantValues(rootValueId, List.of(nodeId))` used instead.

this removes duplicated code and also, from what I can tell, simplifies the call sites that right now distinguish between the two

src/main/java/io/hyperfoil/tools/h5m/svc/ValueService.java
```diff
@@ -688,6 +688,53 @@ WITH RECURSIVE sourceRecursive (v_id) AS (
...
+                
+                        WITH RECURSIVE sourceRecursive (v_id) AS (
+                    SELECT ve.child_id from value_edge ve where ve.parent_id = :rootId
+                    UNION ALL
+                    SELECT ve.child_id from value_edge ve JOIN sourceRecursive sr
+                    ON ve.parent_id = sr.v_id
+                )
+                SELECT distinct v.id FROM value v JOIN sourceRecursive sr ON v.id = sr.v_id WHERE v.node_id in :nodeId
+                """
+        ).setParameter("rootId", rootValueId).setParameter("nodeId",nodeId).getResultList();
```
barreiro: use `.getResultstream().map(Number::longValue).toList()` to skip the intermediate list

src/main/java/io/hyperfoil/tools/h5m/rest/ValueResource.java
```diff
@@ -45,16 +45,25 @@ public JqValue getValueData(@PathParam("id") Long id) {
...
+
     @GET
     @Path("{id}/descendants")
     @PermitAll
     @Operation(description = "Get descendant values of a value. Use ?detection=true to filter to detection node values only.")
     public List<Value> getDescendants(
             @PathParam("id") Long id,
-            @QueryParam("detection") @DefaultValue("false") boolean detectionOnly) {
+            @QueryParam("detection") @DefaultValue("false") boolean detectionOnly,
+            @QueryParam("node") List<Long> nodes) {
```
barreiro, line 56: I would like to see a better description on how the endpoint operates and what the parameters do

also, the precedence of the `detection` parameter should also be documented

src/main/java/io/hyperfoil/tools/h5m/svc/ValueService.java
```diff
@@ -688,6 +688,53 @@ WITH RECURSIVE sourceRecursive (v_id) AS (
...
+        List<Long> longIds = ids.stream().map(Number::longValue).toList();
+        CycleAvoidingContext ctx = new CycleAvoidingContext();
+        return em.unwrap(Session.class)
+                .findMultiple(ValueEntity.class, longIds)
+                .stream().map(e -> apiMapper.toValue(e, ctx))
+                .toList();
+    }
+    public List<Value> getDescendantValues(long rootValueId,List<Long> nodeId) {
+        List<Number> ids = em.createNativeQuery(
```
barreiro: I think there should be a check that the `rootValueId` exists and throw an error if the value does not exist. 

otherwise the caller get an empty list both an invalid rootValueId and a Value without descendants in the specified nodes. the caller won't be able to distinguish between the two cases.

src/main/java/io/hyperfoil/tools/h5m/rest/ValueResource.java
```diff
@@ -45,16 +45,25 @@ public JqValue getValueData(@PathParam("id") Long id) {
         return data;
     }
 
+
```
barreiro, line 48:
```suggestion
```

src/main/java/io/hyperfoil/tools/h5m/svc/ValueService.java
```diff
@@ -688,6 +688,53 @@ WITH RECURSIVE sourceRecursive (v_id) AS (
...
+                """
+                
+                        WITH RECURSIVE sourceRecursive (v_id) AS (
```
barreiro:
```suggestion
                WITH RECURSIVE sourceRecursive (v_id) AS (
```

barreiro (changes requested):

src/main/java/io/hyperfoil/tools/h5m/rest/ValueResource.java
willr3: updated based on our conversation

src/main/java/io/hyperfoil/tools/h5m/svc/ValueService.java
willr3: That functionality sounds good but I don't think it exists on the other api calls so I believe it is out of scope here. Do you see that behavior on the existing api?

src/main/java/io/hyperfoil/tools/h5m/svc/ValueService.java
willr3: updated

src/main/java/io/hyperfoil/tools/h5m/svc/ValueService.java
willr3: removed

src/main/java/io/hyperfoil/tools/h5m/svc/ValueService.java
willr3: this should be fixed, auto formatting is a pain

barreiro (approved):

PR #182, opened 2026-07-05
Issue 173 bytea jdbctype

stalep:

PR #248, opened 2026-08-08
Import performance optimizations: 2.5x throughput improvement

stalep:
Three optimizations for the bulk import pipeline, verified with async-profiler CPU and allocation profiling on test 339 (rhivos-perf-comprehensive, 1,327 runs, 129 nodes):

1. Cache ValueEntity.sources collection in 2LC (ValueEntity.java)
   - Add @Cache(READ_ONLY) on the sources @ManyToMany collection
   - Eliminates N+1 collection loading when entities are served from the Hibernate second-level cache
   - Restore @Mutability(Immutability.class) on data field, accidentally removed in PR #182. Tells Hibernate the JqValue is immutable, reducing deep-copy overhead during cache entry construction (~6% throughput improvement)

2. Batch JPQL for source value loading in WorkService (WorkService.java)
   - Replace N separate em.find() + Hibernate.initialize(data) calls with a single JPQL query that eagerly loads source values with their data and sources collections in one round-trip
   - The 2LC does not cache @Basic(LAZY) BYTEA properties for entities with associations (filed as HHH-20773), so em.find() cache hits still triggered a DB round-trip for every lazy data access
   - The JPQL query bypasses this limitation entirely

3. Pre-compute ancestor node IDs for Work dependency checks (Work.java)
   - Add precomputeAncestors() that collects all transitive ancestor node IDs into a HashSet while the Hibernate session is open
   - dependsOn() uses O(1) Set.contains() instead of O(n) graph traversal through NodeEntity.dependsOn() which triggered Hibernate bytecode interceptor overhead on every getId() call
   - Called in WorkService.create() replacing the work.dependsOn(work) hack that only initialized lazy sources without caching results
   - Eliminates ~16% CPU from KahnDagSort during WorkQueue.addWorks()

Benchmark: test 339, 100 runs, batch=5, 8GB heap, 4 core / 10 max pool
  Before: 176 values/sec (baseline main branch)
  After:  434 values/sec (+147%)
  Steady-state peaks: 500-535 values/sec

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/entity/work/Work.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/entity/work/Work.java
@@ -23,6 +23,12 @@ public class Work implements Runnable, Comparable<Work>{
 
     private Set<NodeEntity> activeNodes;
 
+    // Cached set of all transitive ancestor node IDs for activeNodes.
+    // Populated by precomputeAncestors() while the Hibernate session is open,
+    // then used by dependsOn() for O(1) lookups during WorkQueue sorting
+    // (which runs in afterCompletion, outside the session).
+    private Set<Long> ancestorNodeIds;
+
     /*
      * If the work should be performed after work for any dependent Nodes regardless of Values
      */
@@ -61,6 +67,7 @@ public class Work implements Runnable, Comparable<Work>{
 
     public void setActiveNodes(Set<NodeEntity> activeNodes) {
         this.activeNodes = activeNodes;
+        this.ancestorNodeIds = null; // invalidate cache — new active nodes need fresh ancestors
         if(activeNodes.stream().anyMatch(node -> node instanceof StdDevAnomaly || node instanceof EDivisive)){
             this.cumulative = true;
         }else{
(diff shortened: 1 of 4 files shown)


src/main/java/io/hyperfoil/tools/h5m/entity/work/Work.java
```diff
@@ -69,6 +75,29 @@ public void setActiveNodes(Set<NodeEntity> activeNodes) {
...
     public List<Long> getSourceValueIds(){return sourceValueIds;}
 
+    /**
+     * Pre-computes the transitive ancestor node IDs for all active nodes.
+     * Must be called while the Hibernate session is open (sources are lazy).
+     * After this, dependsOn() uses O(1) Set.contains() instead of traversing
+     * the NodeEntity.sources graph.
+     */
+    public void precomputeAncestors() {
+        if (activeNodes == null || activeNodes.isEmpty()) return;
```
barreiro, line 86: the `setActiveNodes()` method right now it's not called anywhere. it should invalidate the `ancestorNodeIds` cache (or maybe removed altogether to reinforce that assumption that the node tree does not change for the `Work`).

src/main/java/io/hyperfoil/tools/h5m/entity/work/Work.java
```diff
@@ -69,6 +75,29 @@ public void setActiveNodes(Set<NodeEntity> activeNodes) {
...
+        ancestorNodeIds = new HashSet<>();
+        for (NodeEntity activeNode : activeNodes) {
+            if (activeNode.sources == null) continue;
+            Queue<NodeEntity> queue = new ArrayDeque<>(activeNode.sources);
```
barreiro: the `queue` should be created outside the loop and `addAll()` called in each iteration.

src/main/java/io/hyperfoil/tools/h5m/svc/WorkService.java
```diff
@@ -308,16 +306,21 @@ public void execute(Work w){
...
+            // query. em.find() + Hibernate.initialize(data) would trigger N
+            // separate lazy-load SELECTs because the 2LC does not cache lazy
+            // BYTEA properties. This JPQL query loads all values with their
+            // data column in one round-trip.
+            List<ValueEntity> sourceValues;
+            List<Long> sourceIds = w.getSourceValueIds();
+            if (sourceIds == null || sourceIds.isEmpty()) {
+                sourceValues = List.of();
+            } else {
+                sourceValues = em.createQuery(
```
barreiro, line 323: This code does not do what the comment says it does: `data` is still lazy loaded in the `LEFT JOIN FETCH`.

The fix is a JPA Entity Graph — it tells Hibernate "for this query, treat these attributes as eager":
```java
EntityGraph<ValueEntity> graph = em.createEntityGraph(ValueEntity.class);
graph.addAttributeNodes("data", "sources");

sourceValues = em.createQuery("SELECT v FROM value v WHERE v.id IN :ids", ValueEntity.class)
     .setHint("jakarta.persistence.fetchgraph", graph)
     .setParameter("ids", sourceIds)
     .getResultList();
```

barreiro (changes requested):

barreiro (approved):

PR #250, opened 2026-08-08
add NodeConfiguration interface for typed detection config

barreiro:
Move detection config records into `api/node` package behind a sealed `NodeConfiguration` interface with JSON-B type deduction deserializer.

Entity constructors now accept config records directly, removing the Map/record dual-dispatch conversion methods from `NodeService`.

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/api/node/NodeConfiguration.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/api/node/NodeConfiguration.java
@@ -1,0 +1,37 @@
+package io.hyperfoil.tools.h5m.api.node;
+
+import jakarta.json.Json;
+import jakarta.json.JsonObject;
+import jakarta.json.bind.annotation.JsonbTypeDeserializer;
+import jakarta.json.bind.serializer.DeserializationContext;
+import jakarta.json.bind.serializer.JsonbDeserializer;
+import jakarta.json.stream.JsonParser;
+import jakarta.json.stream.JsonParserFactory;
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+
+import java.lang.reflect.Type;
+
+@JsonbTypeDeserializer(NodeConfiguration.Deserializer.class)
+@Schema(oneOf = {FixedThresholdConfig.class, RelativeDifferenceConfig.class, StdDevAnomalyConfig.class, EDivisiveConfig.class})
+public sealed interface NodeConfiguration permits FixedThresholdConfig, RelativeDifferenceConfig, StdDevAnomalyConfig, EDivisiveConfig {
+
+    class Deserializer implements JsonbDeserializer<NodeConfiguration> {
+        private static final JsonParserFactory PARSER_FACTORY = Json.createParserFactory(null);
+
+        @Override
+        public NodeConfiguration deserialize(JsonParser parser, DeserializationContext ctx, Type rtType) {
+            JsonObject obj = parser.getObject();
+            Class<? extends NodeConfiguration> type;
+            if (obj.containsKey("windowLen")) {
+                type = EDivisiveConfig.class;
+            } else if (obj.containsKey("windowSize")) {
+                type = StdDevAnomalyConfig.class;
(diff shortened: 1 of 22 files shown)


src/main/java/io/hyperfoil/tools/h5m/api/node/NodeConfiguration.java
```diff
@@ -0,0 +1,34 @@
...
+            if (json.containsKey("windowLen")) {
+                type = EDivisiveConfig.class;
+            } else if (json.containsKey("windowSize")) {
+                type = StdDevAnomalyConfig.class;
+            } else if (json.containsKey("threshold")) {
+                type = RelativeDifferenceConfig.class;
+            } else {
+                type = FixedThresholdConfig.class;
+            }
+            return ctx.deserialize(type, jakarta.json.Json.createParser(new StringReader(json.toString())));
```
stalep: Why re-parse the json here?

src/main/java/io/hyperfoil/tools/h5m/api/node/NodeConfiguration.java
```diff
@@ -0,0 +1,34 @@
...
+
+@JsonbTypeDeserializer(NodeConfiguration.Deserializer.class)
+@Schema(oneOf = {FixedThresholdConfig.class, RelativeDifferenceConfig.class, StdDevAnomalyConfig.class, EDivisiveConfig.class})
+public sealed interface NodeConfiguration permits FixedThresholdConfig, RelativeDifferenceConfig, StdDevAnomalyConfig, EDivisiveConfig {
+
+    class Deserializer implements JsonbDeserializer<NodeConfiguration> {
+        @Override
+        public NodeConfiguration deserialize(JsonParser parser, DeserializationContext ctx, Type rtType) {
+            JsonObject json = parser.getObject();
+            Class<? extends NodeConfiguration> type;
```
stalep, line 24: This if/else chain is a bit "brittle" but it works :)

src/main/java/io/hyperfoil/tools/h5m/rest/NodeResource.java
```diff
@@ -64,7 +66,7 @@ public Node createConfigured(
             @QueryParam("groupId") @NotNull Long groupId,
             @QueryParam("type") @NotNull NodeType type,
             @QueryParam("sources") @NotNull @NotEmpty List<Long> sources,
-            Object configuration) {
+            @RequestBody(required = false) NodeConfiguration configuration) {
```
stalep, line 69: Do we later check the possibility of configuration being null with this change? If not we might get a NPE

stalep (commented):
Looks good, only had some minor comments

src/main/java/io/hyperfoil/tools/h5m/rest/NodeResource.java
barreiro: I agree, but the thing is that `FingerprintNode` at the moment is created on this endpoint and it doesn't have a configuration !!

src/main/java/io/hyperfoil/tools/h5m/api/node/NodeConfiguration.java
barreiro: using Jackson instead of JSON-B this entire deserializer would be unnecessary: `@JsonTypeInfo(use = JsonTypeInfo.Id.DEDUCTION)` on the sealed interface gives automatic type deduction out of the box

might be worth considering the switch

src/main/java/io/hyperfoil/tools/h5m/api/node/NodeConfiguration.java
barreiro: it's a JSON-B API limitation. after consuming the parser to inspect keys for type deduction, we have to create a new parser from the already-parsed object.

we could construct the records manually from the `JsonObject` to avoid re-parsing, but then the deserializer duplicates knowledge of every record's fields and must be updated whenever a config record changes.

I remove the string serialization and use `JsonParserFactory.createParser(JsonObject)` since a new parser from the JSON object just walks the already in-memory tree.

PR #254, opened 2026-08-17
consolidate ProcessingStatus and RecalculationStatus into Processing

barreiro:
Summary:
- Remove `ProcessingType`, `ProcessingState`, `ProcessingStatus`, `RecalculationStatus`, `Upload`, `UploadTracker`, `RecalculationTracker`, and `RecalculationService` — their responsibilities are consolidated into `Processing` and`ProcessingService`
- Remove the concept of "upload" from the backend. Replaced with "value" / "root value" for the data and  "ingestion" for the operation of calculating the values (like "recalculation" is the operation on a node change). 
- Replace `ProcessingTrackerEntity` with `ProcessingEntity`, splitting the polymorphic `referenceId` field into explicit `nodeId` and `valueId` fields — type is now derived (`isIngestion()` / `isRecalculation()`) instead of stored in an enum
- Remove `CompletionFuture` from the API. users can now await for value calculation on `ProcessingService.await*` methods.
- Fix `completeIngestion` query to filter on `completed = false`, preventing a race where a finished ingestion record could be matched instead of the active one.
- Remove `CompletionFuture` from the API. users can now await for value calculation on `ProcessingService.await*` methods.
- `ProcessingService` keeps internal mutable state on `ActivityTracker` objects that tests can access to monitor ephemeral value eviction (the `afterCleanup` future).
- Move root value creation from `FolderService.upload` to `ValueService.createRootValue`, with `ProcessingService` owning the full tracker lifecycle for both ingestion and recalculation
- Move node recalculation from `FolderService` to `ProcessingService`
- REST paths and CLI commands keep "upload" naming for backwards compatibility

Test plan:
- [ ] Full test suite passes on default profile (699 tests)
- [ ] Full test suite passes on SQLite profile (698 tests)
- [ ] REST upload endpoint unchanged (backwards compatible)
- [ ] CLI upload/status commands unchanged (backwards compatible)

Points for discussion:
- Is "ingestion" the right name for the operation ?! I find "calculation" too similar to "recalculation" and could create issues and confusion.
- Should the CLI commands and REST endpoints for upload and node recalculation be changed as well, to better match the backend ??


willr3:
I agree with the goal to unify the processing trigged by uploading data or changing the node graph but I think the implementation details here are not yet ready for merging. I'm concerned about removing the CompletableFuture an requiring active polling. Furthermore, the active polling implementation will block a thread per poll request. That is basically a DDOS vector. I think the ProcesssingResource can have an endpoint that waits with timeout using a CompletableFuture so that we do not have to park a thread per timeout request. Retaining CompletableFuture also helps avoid a background thread per cleanup step (e.g. ephemeral data)

src/main/java/io/hyperfoil/tools/h5m/rest/NodeResource.java
```diff
@@ -75,50 +79,38 @@ public Node createConfigured(
...
-        // Trigger recalculation if the operation changed. recalculateNode()
-        // opens a requiringNew transaction that only reads the node's group
-        // and folder structure — not the operation. The actual operation is
-        // read later by async Work executors, after this @Transactional
-        // method returns and the update is committed.
-        RecalculationStatus status = null;
-        if (operationChanged && existing.group != null) {
-            status = folderService.recalculateNode(id).toStatus();
+        if (operation != null && !Objects.equals(existing.operation, operation) && existing.group != null) {
+            processingService.recalculateNode(id);
```
stalep: Will this ever be true since we set existing.operation = operation on line 98?

barreiro: good point. fixed and also clarified the case for null vs. empty `name` and `operation` parameters

src/main/java/io/hyperfoil/tools/h5m/svc/ProcessingService.java
```diff
@@ -48,14 +69,177 @@ public class ProcessingService {
...
+            tracker.fail(t);
+        }
+    }
+
+    // --- Status queries ---
+
+    @Override
+    @Transactional
+    public Processing getIngestionStatus(long rootValueId) {
+        ActivityTracker tracker = byRootValueId.get(rootValueId);
```
willr3: abstraction violation? Should this use `getByRootValueId`?

willr3:
> I agree with the goal to unify the processing trigged by uploading data or changing the node graph but I think the implementation details here are not yet ready for merging. I'm concerned about removing the CompletableFuture an requiring active polling. Furthermore, the active polling implementation will block a thread per poll request. That is basically a DDOS vector. I think the ProcesssingResource can have an endpoint that waits with timeout using a CompletableFuture so that we do not have to park a thread per timeout request. Retaining CompletableFuture also helps avoid a background thread per cleanup step (e.g. ephemeral data)

To clarify (now that I'm more level headed) The proposed method for waiting for processing to complete is via ProcessingServiceInterface.awaitRecalculation / awaitIngestion which are blocking calls. I was concerned about them being added to the ProcessingResource because it would block a thread per call to the method. Thankfully they are not part of ProcessingResource and I quite like the direction this change is taking h5m. I do think we should removing the blocking methods from ProcessingServiceInterface in favor of returning a completablefuture and allowing the calling thread to block if desired but that is something we can discuss in a separate issue.

willr3:
```
WARN  [io.smallrye.openapi.runtime.scanner.spi] (build-23) SROAP07903: Duplicate operationId: getRecalculationStatus produced by Class: io.hyperfoil.tools.h5m.rest.ProcessingResource, Method: io.hyperfoil.tools.h5m.api.Processing getRecalculationStatus(long nodeId) and Class: io.hyperfoil.tools.h5m.rest.FolderResource, Method: io.hyperfoil.tools.h5m.api.Processing getRecalculationStatus(long nodeId)
```
I think we can drop the endpoint from the FolderResource?

willr3:
Tests pass when I run them locally so I am ready to merge once we resolve the Duplicate operationId (which logs every single time h5m runs).

barreiro:
> I think we can drop the endpoint from the FolderResource?

yeah, forgot to remove it !!

I rebased, so now all tests pass

PR #257, opened 2026-08-17
add exponential backoff sleep on SQLite retries

barreiro:
attempt to fix some tests that are flaky on CI when running on SQLite


stalep:
I think it looks good, but it changes the Log to debug instead of info. I think info might be better here so we can more easily find issues in CI/locally?

barreiro:
> I think it looks good, but it changes the Log to debug instead of info. I think info might be better here so we can more easily find issues in CI/locally?

agree. in a second though, I think it can be helpful to see the reties on the CI logs to understand if the default values are appropriate !!

barreiro:
also fixed the detection of `PessimisticLockException` that sometimes was being lost / converted in `QuarkusTransaction.requiringNew().call()`

PR #263, opened 2026-08-17
Use FingerprintNode instead of JqNode for fingerprint sources in tests

### LINKED ISSUE #252: ChangeDetectionTest and StdDevAnomalyTest use JqNode instead of FingerprintNode for fingerprint sources
## Problem

`ChangeDetectionTest` and `StdDevAnomalyTest` create fingerprint nodes as `JqNode` instances instead of proper `FingerprintNode` entities:

```java
// ChangeDetectionTest.java:53
NodeEntity fingerprintNode = new JqNode("fingerprint", ".fingerprint");
ft.setNodes(fingerprintNode, rootNode, rangeNode);

// StdDevAnomalyTest.java
NodeEntity fingerprintNode = new JqNode("fingerprint", ".fingerprint", splitNode);
sd.setNodes(fingerprintNode, splitNode, rangeNode, domainNode);
```

This works because `setNodes()` accepts `NodeEntity` (base type), and for single-source fingerprints the extracted value happens to be equivalent to what `FingerprintNode.calculateFpValues()` would produce.

However, it is semantically incorrect — `FingerprintNode` has specific behavior in `calculateFpValues()` that combines multiple source values into a single `{"key1": val1, "key2": val2}` JSON object. Using `JqNode` bypasses this logic and would produce different results for multi-source fingerprints.

## Suggested Fix

Update both tests to use `FingerprintNode` wrapping a JQ extractor:

```java
// Before:
NodeEntity fingerprintNode = new JqNode("fingerprint", ".fingerprint");
fingerprintNode.persist();

// After:
JqNode fpExtractor = new JqNode("fpExtractor", ".fingerprint");
fpExtractor.persist();
FingerprintNode fingerprintNode = new FingerprintNode("fingerprint", "", List.of(fpExtractor));
fingerprintNode.persist();
```

This matches how fingerprint nodes are created in:
- `NodeServiceTest` — uses `new FingerprintNode("fp", "fp", List.of(sourceA, sourceB))`
- `EDivisiveTest` — uses `new FingerprintNode("fp", "", List.of(fpSource))`
- `RecalculateTest` — uses `nodeService.createConfigured("_fp", ..., FINGERPRINT, ...)`
- CLI detection commands (`AddFixedThreshold`, etc.) — internally create `FingerprintNode` via `createConfigured`

## Files to modify

- `src/test/java/io/hyperfoil/tools/h5m/svc/ChangeDetectionTest.java` (line 53)
- `src/test/java/io/hyperfoil/tools/h5m/svc/StdDevAnomalyTest.java`

stalep:
Replace plain JqNode with a JqNode extractor wrapped in a FingerprintNode in ChangeDetectionTest, StdDevAnomalyTest, and EDivisiveTest. This matches the correct two-node pattern used elsewhere (e.g., EDivisiveTest.setupAndUpload) where the extractor feeds into a FingerprintNode that produces structured sorted-key JSON fingerprints via calculateFpValues().

close #252

### DIFF
--- a/src/test/java/io/hyperfoil/tools/h5m/svc/ChangeDetectionTest.java
+++ b/src/test/java/io/hyperfoil/tools/h5m/svc/ChangeDetectionTest.java
@@ -4,6 +4,7 @@ import io.hyperfoil.tools.jjq.value.*;
 import io.hyperfoil.tools.h5m.FreshDb;
 import io.hyperfoil.tools.h5m.entity.NodeEntity;
 import io.hyperfoil.tools.h5m.entity.ValueEntity;
+import io.hyperfoil.tools.h5m.entity.node.FingerprintNode;
 import io.hyperfoil.tools.h5m.entity.node.FixedThreshold;
 import io.hyperfoil.tools.h5m.entity.node.JqNode;
 import io.hyperfoil.tools.h5m.entity.node.RootNode;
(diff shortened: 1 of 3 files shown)


willr3:
Please see the comments on  issue #252. I do not believe this PR should be merged or considered until we agree there is an issue that needs to be resolved. In either case, the code change proposed here fails to produce a correct value graph and fails to accomplish the stated goals of #252.

PR #264, opened 2026-08-18
Upgrade jjq to 0.1.9 and fix JS proxy mutation for legacy Horreum fun…

stalep:
…ctions

Upgrade jjq from 0.1.8 to 0.1.9, adding JqValue.estimatedSizeInBytes() and JqValues.serializeToByteOutput() for future weight-based 2LC cache eviction.

Fix ProxyJqObject and ProxyJqArray to support mutation through the GraalVM proxy chain. Legacy Horreum JS functions mutate nested objects (e.g., value["pcp_time_series"][i].elapsed_time = x), but the previous implementation threw UnsupportedOperationException, silently discarding mutations and causing fillInStackTrace overhead on every attempt.

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/pasted/ProxyJqArray.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/pasted/ProxyJqArray.java
@@ -1,46 +1,131 @@
 package io.hyperfoil.tools.h5m.pasted;
 
 import io.hyperfoil.tools.jjq.value.JqArray;
+import io.hyperfoil.tools.jjq.value.JqNull;
 import io.hyperfoil.tools.jjq.value.JqValue;
 import org.graalvm.polyglot.Value;
 import org.graalvm.polyglot.proxy.ProxyArray;
 
+import java.util.ArrayList;
+import java.util.List;
+
 /**
  * GraalVM ProxyArray backed by JqArray.
- * Replaces ProxyJacksonArray — eliminates the JqValue→JsonNode conversion
- * for JavaScript interop.
+ * <p>
+ * Supports all standard JavaScript array operations. On first mutation
+ * ({@link #set} or {@link #remove}), the backing switches from the
+ * immutable {@link JqArray} to a mutable {@link ArrayList} that supports
+ * index-based set (including array growth), element removal with index
+ * shifting, and nested proxy caching for mutation propagation.
+ * <p>
+ * {@link #getJqArray()} rebuilds an immutable JqArray from the current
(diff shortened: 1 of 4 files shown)


src/main/java/io/hyperfoil/tools/h5m/pasted/ProxyJqArray.java
```diff
@@ -1,46 +1,120 @@
...
+            overlay = new HashMap<>();
+        }
+        JqValue jqValue = Util.convertToJqValue(value);
+        overlay.put((int) index, jqValue != null ? jqValue : JqNull.NULL);
     }
 
     @Override
     public boolean remove(long index){
-        throw new UnsupportedOperationException("JqArray is immutable");
+        throw new UnsupportedOperationException("JqArray does not support remove");
```
willr3: Why is this remaining unsupported? Horreum and h5m do not advertise that the javascript engine is an incomplete implementation of ecma so I do not think we should intentionally disable core functionality.

src/main/java/io/hyperfoil/tools/h5m/pasted/ProxyJqArray.java
```diff
@@ -1,46 +1,120 @@
...
     }
 
     @Override
-    public long getSize(){ return node.length(); }
+    public long getSize(){ return original.length(); }
```
willr3: What happens if the mutation increases the length of the array? In javascript it is valid to increase an array length with a set operation
```
var foo = [1,2,3]
foo[3]=4
```

src/main/java/io/hyperfoil/tools/h5m/pasted/ProxyJqObject.java
```diff
@@ -30,42 +37,125 @@ public Object execute(Value...args){
...
+            overlay = new LinkedHashMap<>();
+        }
+        JqValue jqValue = Util.convertToJqValue(value);
+        overlay.put(key, jqValue != null ? jqValue : JqNull.NULL);
     }
 
     @Override
     public boolean removeMember(String key) {
-        throw new UnsupportedOperationException("JqObject is immutable");
+        throw new UnsupportedOperationException("JqObject does not support removeMember");
```
willr3: I have the same concern here. I think the lesson should be that we need the proxy to support all valid javascript operations on the underlying object

willr3 (changes requested):
I think leaving any part of the necessary mutation API as unsupported is going to invite the same error conditions to occur again.

PR #270, opened 2026-08-19
fix fingerprint use of mutable fields and detection node filtering

willr3:
The current implementation for fingerprint stores the source node names in the ValueEntity whenever it calculates a new Value. The name is a mutable field and storing it means having to potentially recalculate all fingerprints whenever a source node name changes. Instead, this PR changes fingerprints to using the source node ids as keys (an immutable field) and recreates the `{ [name] : value}` object whenever it is needed for a DetectionNode's fingerprintFilter.

This PR will also update DetectionNode fingerprintFilters whenever the node name changes for the associated FingerprintNode. This will keep the filter in sync with mutable node name. The update uses the same method we currently use to keep JsNode parameter names in sync with name mutations to their source nodes.

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
@@ -170,13 +170,23 @@ public class NodeService implements NodeServiceInterface {
         }else{
             NodeEntity existing = NodeEntity.findById(node.id);
             if(!existing.name.equals(node.name)){
-                List<NodeEntity> toChange = em.createNativeQuery(
-                        "select n.* from node n join node_edge ne on n.id = ne.child_id where ne.parent_id=? and n.type='ecma'"
-                , NodeEntity.class).setParameter(1,node.id).getResultList();
                 Map<String,String> changes = Map.of(existing.name,node.name);
-                for(NodeEntity n : toChange){
-                    n.operation = renameParameters(n.operation, changes);
-                    em.merge(n);
+                //change ecma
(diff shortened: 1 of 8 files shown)


src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
```diff
@@ -170,10 +170,26 @@ public long update(NodeEntity node){
         }else{
             NodeEntity existing = NodeEntity.findById(node.id);
             if(!existing.name.equals(node.name)){
+                Map<String,String> changes = Map.of(existing.name,node.name);
+                //change ecma
+                for(NodeEntity dependent : getDependentNodes(existing)){
+                    if(dependent.type().equals(NodeType.JS)){
+                        dependent.operation = renameParameters(dependent.operation,changes);
```
stalep: Isn't this already done in the query below on line 189-195?
Should we just remove the query below?

stalep (commented):
I think this pr looks good. We probably need to expose a method to fetch the fingerprint names for the cli and web ui.

src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
willr3: Good catch! Yes, the `toChange` collection is no longer valid, I will remove

willr3: Also adding explicit calls to `em.merge` as that was present in the loop for `toChange`

PR #274, opened 2026-08-19
initial authentication infrastructure 

barreiro:
Summary

- Add `quarkus-oidc-proxy` extension to expose a `/q/oidc` discovery endpoint, allowing the frontend OIDC client to discover the identity provider through the backend. Dev profile auto-starts an OIDC server with test users (roles are assigned by h5m on first login).
- Store OIDC `sub` and `iss` claims in `UserEntity` as stable identity. The provisioner and roles augmentor pattern-match on principal type (`JsonWebToken` vs plain principal) to dispatch user lookup accordingly.
- Wire up frontend auth: `AuthorizationContext`/`Provider` injects OIDC tokens into Axios via interceptor, fetches the user role from `/api/user/role`, and exposes `isAdmin`/`isAuthenticated`. `AuthActions` component provides login/logout with token renewal and notifications.

Test plan

- [x] Start dev mode — OIDC server launches automatically with test users
- [x] Login via OIDC — username displays in the header (not "Anonymous")
- [x] First user gets `ADMIN` role, subsequent users get `USER` role
- [x] `AutoProvisioningTest` covers OIDC provisioning with `sub`/`iss` and the username-based path
- [x] API key auth remains functional with OIDC enabled
- [x] Create folder using CLI does not require authentication

### DIFF
--- a/src/main/resources/application.properties
+++ b/src/main/resources/application.properties
@@ -51,6 +51,9 @@ quarkus.banner.enabled=false
 %cli.quarkus.datasource.jdbc.url=jdbc:sqlite:${H5M_PATH:${user.home}/h5m.db}
 %cli.quarkus.datasource.devservices.enabled=false
 %cli.quarkus.quinoa=false
+# disable OIDC
+%cli.quarkus.oidc.tenant-enabled=false
+%cli.quarkus.oidc-proxy.enabled=false
 # CLI uses uber-jar for single-file distribution. Truffle Multi-Release check is disabled since uber-jar merging loses the Multi-Release manifest attribute.
 %cli.quarkus.package.jar.type=uber-jar
 %cli.quarkus.package.jar.add-runner-suffix=false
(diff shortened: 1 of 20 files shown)


src/main/resources/application.properties
```diff
@@ -92,20 +95,21 @@ quarkus.hibernate-orm.log.queries-slower-than-ms=100
...
-h5m.security.enabled=false
+# ApiKey Security
 h5m.api-key.expiration-days=365
 
 # Disable proactive auth so unauthenticated requests can reach @PermitAll endpoints
 quarkus.http.auth.proactive=false
 
-# OIDC - disabled by default (enable via environment for service deployments)
-quarkus.oidc.tenant-enabled=false
-quarkus.oidc.auth-server-url=${OIDC_AUTH_SERVER_URL:}
```
willr3: Does removing `quarkus.oidc.auth-server-url=${OIDC_AUTH_SERVER_URL:}` mean the `quarkus.oidc.auth-server-url` property must be in either a `-D` command line argument or in the `.env` when running `h5m`?

willr3 (commented):
How are we / users going to specify the `auth-server-url`?

src/main/resources/application.properties
barreiro: yes. in a production setting the `quarkus.oidc.auth-server-url` for the provider will be specified, along with `client-id` + `credentials-secret` and others, using one of the several methods available to specify / override properties.

I see no point in having an alias, when the original quarkus properties are available and are well documented.

willr3: Would it be helpful to keep those properties in application.properties (possibly without values)  so we know they must be defined / overridden when deploying h5m?

PR #281, opened 2026-08-21
Standardize CLI verb consistency and restructure admin commands

### LINKED ISSUE #277: admin commands use verb->noun breaking cli consistency
The admin entity reverts to verb-noun named commands which break user consistency. If we are moving users to a entity centric command line then we should not mix conventions.
```
[h5m]$ admin
add-member  create-api-key  create-team  create-user  list-api-keys  list-teams  list-users  revoke-api-key
```

One noun->verb consistent approach could be to un-group the member, team, user, api-key entities.
```
[h5m]$ member add ...
```

### LINKED ISSUE #278: command verbs are not consistent across entities
The previous cli forced verb consistency across operations (create, read, update, delete) by using the verb as the command grouping. We should continue to maintain verb consistency for all CRUD operations even with a noun->verb based cli. At the moment we have some commands using `add` while others users use `create`. 

```
[h5m]$ view
create  list  remove  show  update
[h5m]$ node
add  list  remove  update
```

stalep:
Fix #278: Standardize CRUD verbs across all entity groups. Change 'view create' to 'view add' (with 'create' as alias for backwards compatibility), matching folder/node/notification which all use 'add'.

Fix #277: Restructure admin commands from flat verb-noun naming (create-team, list-users, etc.) to nested entity groups matching the noun-verb pattern used by all other entity commands.

Before:
  admin create-team|create-user|list-teams|list-users|add-member|
        create-api-key|list-api-keys|revoke-api-key

After:
  admin team    → add, list
  admin user    → add, list
  admin member  → add
  admin apikey  → add, list, revoke

This makes admin consistent with the entity-first pattern:
  folder → add, list, remove, upload, values, ...
  node   → add, list, remove, update
  view   → add, list, remove, show, update
  admin team   → add, list
  admin apikey → add, list, revoke

close #277
close #278


willr3:
I thought our proposed permission model would allow anyone to add a User and any Team Member to update a Team. From this I think the only operations that require `admin` permissions are adding and removing a Team. 

We changed to `noun -> verb` because it was meant to improve entity discoverability but this `admin` grouping does the opposite. It hides permission related operations behind a a non-entity noun. If we are going to tell users that "top level commands are entities" then we should not break that consistency.

PR #283, opened 2026-08-21
wire Zod validation from backend annotations to TanStack Form

barreiro:
Summary

- Zod-based form validation: replaced all inline function validators in `CreateNodeModal` with generated Zod schemas (from OpenAPI/Jakarta Bean Validation), removing the manual `validateStep` method entirely
- Validation utilities: added `fieldError` and `groupValidator` in `validation.ts` to bridge Zod schemas with TanStack Form's Standard Schema support and Carbon's error display
- Backend-driven enums: `Filter` (Relative Difference) and `Direction` (StdDev Anomaly) flow as Java enums through OpenAPI to generated Zod schemas, replacing hardcoded frontend option arrays
- Detection config group validation: detection config fields are wrapped in `FormGroup` with generated Zod schemas (`zRelativeDifferenceConfig`, `zEDivisiveConfig`, etc.) as group validators
- Carbon `NumberInput`: switched numeric config fields from `TextInput` to `NumberInput` with proper `number | string` form value types

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/api/node/RelativeDifferenceConfig.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/api/node/RelativeDifferenceConfig.java
@@ -1,7 +1,15 @@
 package io.hyperfoil.tools.h5m.api.node;
 
+import jakarta.validation.constraints.DecimalMin;
+import jakarta.validation.constraints.Positive;
 import org.eclipse.microprofile.openapi.annotations.media.Schema;
 
 @Schema(description = "Configuration for the Relative Difference detection algorithm")
-public record RelativeDifferenceConfig(String filter, double threshold, int window, int minPrevious, String fingerprintFilter) implements NodeConfiguration {
+public record RelativeDifferenceConfig(
+        Filter filter,
+        @DecimalMin("0") double threshold,
+        @Positive int window,
+        @Positive int minPrevious,
+        String fingerprintFilter) implements NodeConfiguration {
+    public enum Filter { MEAN, MIN, MAX }
 }
--- a/src/main/webui/src/app/components/DetectionConfigStep.tsx
+++ b/src/main/webui/src/app/components/DetectionConfigStep.tsx
@@ -2,14 +2,23 @@ import type { EDivisiveConfig, FixedThresholdConfig, NodeConfiguration, NodeType
 import type { FormAsyncValidateOrFn, FormValidateOrFn } from '@tanstack/form-core';
 import type { ReactFormExtendedApi } from '@tanstack/react-form';
 
-import { Checkbox, Select, SelectItem, Stack, TextInput } from '@carbon/react';
+import { fieldError, groupValidator } from '@app/validation.ts';
+import { Checkbox, NumberInput, Select, SelectItem, Stack } from '@carbon/react';
+import { zDirection, zEDivisiveConfig, zFilter, zFixedThresholdConfig, zRelativeDifferenceConfig, zStdDevAnomalyConfig } from '@client/zod.gen.ts';
+import { z, type ZodType } from 'zod';
 
 import type { FormValues } from './CreateNodeModal';
 
-export const DETECTION_SOURCES: { name: 'srcFingerprint' | 'srcGroupBy' | 'srcRange'; label: string; allowedTypes: NodeType[] }[] = [
-  { name: 'srcFingerprint', label: 'Fingerprint node', allowedTypes: ['FINGERPRINT'] },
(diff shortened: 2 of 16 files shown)


src/main/webui/src/app/components/DetectionConfigStep.tsx
```diff
@@ -2,14 +2,23 @@ import type { EDivisiveConfig, FixedThresholdConfig, NodeConfiguration, NodeType
...
-  { name: 'srcRange', label: 'Range node', allowedTypes: ['JQ', 'JS', 'JSONATA', 'SPLIT'] },
+export const DETECTION_SOURCES: {
+  name: 'srcFingerprint' | 'srcGroupBy' | 'srcRange';
+  label: string;
+  allowedTypes: NodeType[];
+  // frontend-only: the API takes a flat sources array, not individual source-role parameters, so per-select validation can't flow from the backend
+  schema: ZodType<string, string>;
+}[] = [
+  { name: 'srcFingerprint', label: 'Fingerprint node', allowedTypes: ['FINGERPRINT'], schema: z.string().min(1) },
+  { name: 'srcGroupBy', label: 'GroupBy node', allowedTypes: ['ROOT', 'JQ', 'JS', 'JSONATA', 'SPLIT'], schema: z.string().min(1) },
```
Arjun-Parmani: Why ROOT is added in the GroupBy nodes?

src/test/java/io/hyperfoil/tools/h5m/svc/NodeServiceTest.java
```diff
@@ -803,7 +804,7 @@ public void calculateRelativeDifference_defaultFilter() throws SystemException,
         assertEquals(1,found.size());
 
         assertEquals(RelativeDifference.DEFAULT_FILTER, relDifference.getFilter(),
-                "getFilter() should return DEFAULT_FILTER ('mean') when no filter is configured");
+                "getFilter() should return default (MEAN) when no filter is configured");
```
Arjun-Parmani: The PR has introduced new annotations  @Min(), @DecimalMin(), @Positive, @UniqueElements, @NotEmpty, and @Value. There should be unit test cases covering these edge cases I suppose. Asserting that it fails if the min decimal is negative, sources are not unique etc...

src/main/java/io/hyperfoil/tools/h5m/api/node/RelativeDifferenceConfig.java
```diff
@@ -1,7 +1,15 @@
...
 
 @Schema(description = "Configuration for the Relative Difference detection algorithm")
-public record RelativeDifferenceConfig(String filter, double threshold, int window, int minPrevious, String fingerprintFilter) implements NodeConfiguration {
+public record RelativeDifferenceConfig(
+        Filter filter,
+        @DecimalMin("0") double threshold,
+        @Positive int window,
+        @Positive int minPrevious,
+        String fingerprintFilter) implements NodeConfiguration {
+    public enum Filter { MEAN, MIN, MAX }
```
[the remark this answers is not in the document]
barreiro: Median never existed on the backend. That's exactly why we want the types / validation to flow from the backend to the frontend 😉

src/main/webui/src/app/components/DetectionConfigStep.tsx
barreiro: `ROOT` is allowed as a `groupBy` source because it acts as the default series identifier when no dedicated label-extraction node exists in the pipeline. The root node outputs the full uploaded document, so grouping by it means "each distinct payload shape is its own series."

The CLI already defaults to `ROOT` when no explicit `groupBy` node is found (see AddDetectionNode.java`), so the UI should allow the same choice.

It doesn't make sense for `srcRange` (which needs a numeric scalar) or `srcFingerprint` (which needs a fingerprint node), but for `groupBy` it's valid and often the simplest option.

src/test/java/io/hyperfoil/tools/h5m/svc/NodeServiceTest.java
barreiro: added test cases

barreiro:
> Why FixedThresholdConfig has no constraints? I could do something like - min 50 and max 0 and it still creates the detection node.

FixedThreshold doesn't have constraints because all its fields are genuinely unconstrained: `min` and `max` can be any value (including negative), either or both can be null (one-sided thresholds), and the booleans are just inclusive/exclusive flags. There's nothing to validate on individual fields. 

Added cross-field validation `min < max` when both are set. That's a semantic check, not something bean validation annotations are well suited for.

PR #286, opened 2026-08-24
Fix detection nodes without domain ordering (issue #284)

### LINKED ISSUE #284: Legacy import creates detection nodes without domain ordering, producing zero change detection results
## Problem

When importing tests from Horreum via `legacy load-tests`, RelativeDifference and EDivisive detection nodes are created without a domain node:

```java
// LoadLegacyTests.java line 582
difference.setNodes(fingerprintNode, groupBy, variableNode, null);
```

In `calculateRelativeDifferenceValues()`, the entire detection logic is guarded by:
```java
if (relDiff.getDomainNode() != null) {
    // ... all detection logic ...
}
```

When domain is null, the method returns an empty list. Similarly, `calculateEDivisiveValues()` logs an error and returns empty when domain is null.

**No change detection is ever performed on imported tests.**

## Evidence

After importing `quarkus-startstop` (100 runs, 380 start time values, 18 RD nodes with proper thresholds and window/minPrevious configuration):
- 0 detection values produced
- All 18 RD nodes have only 3 sources (fingerprint, groupBy, range) — no domain node
- The detection algorithm skips the entire body due to null domain check

## Root Cause

In Horreum, change detection orders datapoints by `run.start` timestamp implicitly. The `run.start` and `run.stop` fields are `@NotNull @JsonProperty(required = true)` in Horreum's API — every run always has timestamps. h5m requires an explicit domain node for ordering but the import doesn't create one.

Additionally, `load-runs` imports runs in `ORDER BY id DESC` (reverse chronological order). Even if detection were working, the `ValueEntity.created_at` timestamps would be in reverse order, producing incorrect detection results.

## Impact

All Horreum tests with RelativeDifference or EDivisive change detection produce zero detections after import to h5m. This blocks the Horreum → h5m migration path — users expect change detection to work the same way after migration.

### Approach A: Inject domain ordering during import (recommended)

Since Horreum always has `run.start` available:

1. **`load-runs`**: Add `start` to the SQL query (`select id, data, start from run ...`), change ordering to `ORDER BY id ASC` for chronological import, and inject `_run_start` (epoch millis) into the uploaded JSON:
   ```java
   Timestamp start = rs.getTimestamp("start");
   JqObject augmented = ((JqObject) data).with("_run_start", JqNumber.of(start.getTime()));
   valueService.createRootValue(folder.id(), augmented);
   ```

2. **`load-tests`**: Create a `_run_start` JQ domain node and wire it as the 4th source for all RD and ED detection nodes:
   ```java
   NodeEntity domainNode = new JqNode("_run_start", "._run_start", root);
   difference.setNodes(fingerprintNode, groupBy, variableNode, domainNode);
   ```

**Pros**: No changes to detection algorithms, clean explicit domain ordering, the `_run_start` value is queryable and chartable.

**Cons**: Modifies the uploaded data by adding a field. Only solves the import path, not direct h5m uploads without timestamps.

### Approach B: Detection algorithms fall back to `created_at` ordering

Make `calculateRelativeDifferenceValues()` and `calculateEDivisiveValues()` work without a domain node by using `ValueEntity.created_at` ordering. The `findMatchingFingerprint()` method already supports `created_at` fallback when domain parameters are null.

**Pros**: No data modification, works for all detection nodes regardless of how they were created, covers both import and direct upload paths.

**Cons**: Requires changes to core detection logic, `created_at` ordering depends on import order (which is currently reverse chronological — must fix `load-runs` regardless), no explicit domain value in detection output for charting.

### Approach C: Hybrid (A + B)

Implement both: inject `_run_start` during import for correct Horreum ordering, AND add `created_at` fallback in detection algorithms as a safety net for cases where no domain is configured.

**Pros**: Complete solution — import path is correct AND direct uploads without timestamps still get detection.

**Cons**: More changes, need to ensure both paths produce consistent results.

### `load-runs` ordering must change regardless

Currently `ORDER BY id DESC` imports newest first. This means:
- `ValueEntity.created_at` is in reverse chronological order
- Approach B's `created_at` fallback would produce incorrect results without this fix
- Users importing with `--limit N` get the N newest runs, which is useful for testing but wrong for chronological ordering

Changing to `ORDER BY id ASC`:
- `created_at` reflects original chronological order
- Users importing with `--limit N` get the N oldest runs (may need `--offset` to skip old runs)
- Or add a `--latest` flag that keeps `DESC` ordering for convenience

### Non-Horreum direct uploads

When users upload directly to h5m (not via legacy import), they may not have a timestamp field. Horreum requires `start`/`stop` on every run, but h5m does not. For these cases:
- Approach A alone is insufficient — only covers the import path
- Approach B or C provides a safety net via `created_at` ordering
- Alternatively, h5m's upload API could accept optional `start`/`stop` parameters (like Horreum) and create a domain node automatically

stalep:
Add created_at fallback to detection algorithms.
calculateRelativeDifferenceValues() now handles null domain node by using created_at-ordered range values via findMatchingFingerprint with null domain parameters.

calculateEDivisiveValues() similarly no longer requires a domain node.

LoadLegacyTests creates domain node from timeline_labels. When importing detection nodes from Horreum, check the test's timeline_labels configuration. If a timeline label maps to an existing node in the graph (e.g., 'Start Time' for rhivos), use it as the domain node for RD and ED detection nodes.

LoadLegacyRuns imports in chronological order.
Changed ORDER BY id DESC to ORDER BY id ASC.

close #284


willr3 (changes requested):
Please include tests that verify `createX` produce correct values and respect the algorthmic bounds (e.g. minPrevous and window).
Please include tests that verify the correct change value is calculated for the algorithm so that we have defense against changes to dependencies (e.g. `findMatchingFingerprint`)
The tests do not need to test out of order like we do for domain based ordering because that condition is not possible for 

Is it possible to combine the logic that creates the `List<Double> converted`, detects changes, and creates values? That appears to be a rather significant section of duplicated code.

willr3:
Please also add a test that createFolder successfully creates the domain node based on timeline_labels

PR #288, opened 2026-08-24
Fix AUTO ephemeral nullifying nodes whose children are KEEP (#285)

### LINKED ISSUE #285: AUTO ephemeral nullifies nodes whose children are KEEP, breaking detection source chain
## Problem

Nodes with `ephemeral = AUTO` are incorrectly nullified when their children are `KEEP` nodes that feed detection nodes. The ephemeral logic only looks one level down -- if any child is a non-analysis node, the parent is treated as intermediate and its data is nullified. It does not consider whether those children are `KEEP` nodes whose data depends on the parent.

## Example: quarkus-startstop

```
rss (AUTO, data=null)
  → startstop rss min kb (KEEP, data=preserved) → rd.startstop rss min kb (rd)
  → startstop rss avg kb (KEEP, data=preserved) → rd.startstop rss avg kb (rd)
  → startstop rss max kb (KEEP, data=preserved) → rd.startstop rss max kb (rd)
```

The `rss` node extracts `[.measurements?[0].data?[]?.RSSkB?]` (an array of RSS values). Its children compute `add / length` (avg), `Math.min` (min), `Math.max` (max) from that array.

Because `rss` has non-analysis children (the JQ min/avg/max nodes), the AUTO ephemeral logic marks it as intermediate and nullifies its data. But those children are `KEEP` nodes that need `rss` data for recalculation. After nullification, any recalculation of the children fails with:

```
Error processing 1487 startstop rss avg kb
values: 1477=929795
Cannot divide NULL by NUMBER
```

## Root Cause

The `nullifyEphemeralData` query in `ValueService`:

```sql
ephemeral = 'AUTO' AND EXISTS (
    SELECT 1 FROM node_edge ne
    JOIN node child ON child.id = ne.child_id
    WHERE ne.parent_id = node.id
    AND child.type NOT IN ANALYSIS_NODES
)
```

This checks: "does this node have any non-analysis children?" If yes, treat as intermediate. But it does not check whether those children are `KEEP` nodes. A node whose non-analysis children are all `KEEP` should NOT be treated as ephemeral -- its data is needed for recalculation.

## Impact

- Detection source chains like `extractor → aggregator (KEEP) → detection (rd)` break when the extractor is nullified
- Recalculation produces errors for `KEEP` nodes whose `AUTO` sources were nullified
- Observed in quarkus-startstop import: `rss`, `start`, `ttfr`, `fd` nodes are all affected

## Proposed Fix

Add a guard to exclude nodes whose non-analysis children are all `KEEP`:

```sql
ephemeral = 'AUTO' AND EXISTS (
    SELECT 1 FROM node_edge ne
    JOIN node child ON child.id = ne.child_id
    WHERE ne.parent_id = node.id
    AND child.type NOT IN ANALYSIS_NODES
    AND child.ephemeral != 'KEEP'
)
```

This means: "only treat as ephemeral if there are non-analysis children that are NOT `KEEP`." If all non-analysis children are `KEEP`, the node's data is preserved for recalculation.

## Related

- #284: Detection nodes without domain ordering (the import that exposed this issue)
- The ephemeral logic in `NodeService.isEphemeral()` has the same pattern and may need the same fix

stalep:
Add 'AND child.ephemeral != KEEP' guard to three SQL queries that determine whether an AUTO node should be treated as ephemeral:

1. ValueService.nullifyEphemeralData() - the query that nullifies value data after upload processing
2. ValueService.markAutoEphemeral() - the query that marks AUTO nodes as DISCARD based on graph structure
3. NodeService.isEphemeral() - the method used by recalculation logic

Previously, an AUTO node with non-analysis children was always treated as intermediate and its data nullified, regardless of whether those children were KEEP nodes.

close #285

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/svc/NodeService.java
@@ -244,6 +244,7 @@ public class NodeService implements NodeServiceInterface {
                             select 1
                             from node_edge ne JOIN node child ON child.id = ne.child_id
                             where ne.parent_id = n.id AND child.type NOT IN ANALYSIS_NODES
+                            AND child.ephemeral != 'KEEP'
                         )
                     ))
              )
--- a/src/main/java/io/hyperfoil/tools/h5m/svc/ValueService.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/svc/ValueService.java
@@ -1041,6 +1041,7 @@ public class ValueService implements ValueServiceInterface {
                       JOIN node child ON child.id = ne.child_id
                       WHERE ne.parent_id = node.id
                         AND child.type NOT IN ANALYSIS_NODES
+                        AND child.ephemeral != 'KEEP'
                     )))
                   AND type NOT IN ROOT_OR_ANALYSIS_NODES
                   -- Protect nodes that are sources of analysis nodes.
@@ -1083,6 +1084,7 @@ public class ValueService implements ValueServiceInterface {
                 JOIN node child ON child.id = ne.child_id
                 WHERE ne.parent_id = node.id
                   AND child.type NOT IN ANALYSIS_NODES
+                  AND child.ephemeral != 'KEEP'
               )
             """.replaceAll("ROOT_OR_ANALYSIS_NODES",NodeService.ROOT_OR_ANALYSIS_NODES)
                         .replaceAll("ANALYSIS_NODES",NodeService.ANALYSIS_NODES))
--- a/src/test/java/io/hyperfoil/tools/h5m/svc/RecalculateTest.java
+++ b/src/test/java/io/hyperfoil/tools/h5m/svc/RecalculateTest.java
@@ -1316,4 +1316,99 @@ public class RecalculateTest extends FreshDb {
         assertNull(deleted, "Tracker for nonexistent node should be deleted during recovery");
         tm.commit();
     }
+
+    /**
+     * Test that AUTO ephemeral nodes whose children are KEEP are not nullified.
+     * <p>
+     * Pipeline: root → extractor (AUTO) → aggregator (KEEP) → detection (rd)
+     * <p>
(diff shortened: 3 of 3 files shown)


willr3 (changes requested):
I am confused, why does having a KEEP child make a node necessary to retain? The purge should not occur until after the child calculates Values.

Would you please include a test that demonstrates the condition this is solving?

willr3 (changes requested):
>Would you please include a test that demonstrates the condition this is solving?

The test `auto_ephemeral_preserves_data_when_child_is_keep` passes when I copy it into main `efc78b5` without the changes proposed in this PR. Did you create this test and verify it fails without the changes proposed in this PR?

Would you please describe the condition you are seeing where a node is calculating a value and the sources' data were `null` due to being ephemeral? 

The test case uses `recalculateNode` but that should automatically queue for recalculation any ephemeral sources along with the target node. If you are seeing a condition where `recalculateNode` fails to produce the correct value then the issue is likely in `NodeService.getEphemeralSources` and not in the logic for which nodes should be ephemeral. 

Please let me know if you are able to create a unit test that demonstrates the failure condition.

willr3:
Right, this test still passes if you comment out the following:
```
        assertEquals(0, extractorNullCount.intValue(),
                "Extractor (AUTO with KEEP child) data should NOT be nullified (#285) — "
                        + "found " + extractorNullCount + " nullified out of " + extractorTotalCount);
```
That tells me this test is verifying the code change in the PR has an effect but it is not verifying there is a failure condition that needs to be fixed by the PR.

Let's take a step back. Can you describe what you see happening in `h5m` that needs to be fixed? It looks like you believe there is a condition where `recalculateNode` is not producing the correct value. Can you create a reproducer where the result of `recalculateNode` has the wrong value? A failing test based on `main` would be fantastic.  We can work together on what needs to be fixed if you can demonstrate an invalid value after `recalculateNode` but I do not see where we are getting invalid values.

PR #292, opened 2026-08-25
initial Team management

barreiro:
Adds end-to-end team management — REST API, services, and web UI — letting admins create, rename, and delete teams and manage membership.

Backend
- `TeamResource` endpoints for team CRUD and member `add`/`remove`/`list`, backed by `TeamService`/`UserService`.
- Team-scoped authorization wiring.
- Use constants for the defined `Role` (like in `H5mRolesAugmentor`, `AuthorizationService`); tests updated.

Web UI
- `TeamsPanel`, `TeamMembersPanel`, and `TeamNameModal` for managing teams and members, wired into routing, header, and auth context.
- Small tweaks on how the documentation `iframe` is displayed. Allow for direct links to docs.

### DIFF
--- a/src/main/java/io/hyperfoil/tools/h5m/entity/UserEntity.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/entity/UserEntity.java
@@ -13,8 +13,8 @@ import jakarta.persistence.ManyToMany;
 import jakarta.persistence.Table;
 import jakarta.persistence.UniqueConstraint;
 
-import java.util.ArrayList;
-import java.util.List;
+import java.util.HashSet;
+import java.util.Set;
 
 @Entity(name = "h5m_user")
 @Table(uniqueConstraints = @UniqueConstraint(columnNames = {"sub", "iss"}))
@@ -35,7 +35,7 @@ public class UserEntity extends PanacheEntityBase {
     public Role role;
 
     @ManyToMany(mappedBy = "members")
-    public List<TeamEntity> teams = new ArrayList<>();
+    public Set<TeamEntity> teams = new HashSet<>();
 
     public UserEntity() {}
 
@@ -53,15 +53,20 @@ public class UserEntity extends PanacheEntityBase {
 
     @Override
     public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
         if (!(o instanceof UserEntity that)) {
             return false;
         }
-        return username.equals(that.username);
+        return id != null && id.equals(that.id);
     }
 
     @Override
     public int hashCode() {
-        return username.hashCode();
+        // Deliberately constant: hashCode must stay stable across the entity lifecycle while this instance sits in a HashSet.
+        // Basing it on id (null until persisted) or username (mutable) would change the hash and turn the entity into an unreachable "ghost" in its collection.
+        return getClass().hashCode();
     }
 
     @Override
--- a/src/main/java/io/hyperfoil/tools/h5m/rest/TeamResource.java
+++ b/src/main/java/io/hyperfoil/tools/h5m/rest/TeamResource.java
@@ -1,0 +1,98 @@
+package io.hyperfoil.tools.h5m.rest;
+
+import io.hyperfoil.tools.h5m.api.Team;
+import io.hyperfoil.tools.h5m.api.User;
+import io.hyperfoil.tools.h5m.api.svc.TeamServiceInterface;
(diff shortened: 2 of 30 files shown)


src/main/java/io/hyperfoil/tools/h5m/entity/UserEntity.java
```diff
@@ -35,7 +35,7 @@ public class UserEntity extends PanacheEntityBase {
     public Role role;
 
     @ManyToMany(mappedBy = "members")
-    public List<TeamEntity> teams = new ArrayList<>();
+    public Set<TeamEntity> teams = new HashSet<>();
```
Arjun-Parmani: Why is List replaced with Set? Furthermore, if a team name can be renamed (via service layer 'renameTeam' ), then the hashcode computes a different value and existing team name is lost and further equals() lookup against that set will silently fail to find it, even though it's technically still sitting in the collection.

src/main/java/io/hyperfoil/tools/h5m/rest/TeamResource.java
```diff
@@ -0,0 +1,98 @@
...
+public class TeamResource {
+
+    @Inject
+    TeamServiceInterface teamService;
+
+    @Inject
+    SecurityIdentity identity;
+
+    @Inject
+    UserService userService;
```
Arjun-Parmani: why not Inject UserServiceInterface?

src/main/java/io/hyperfoil/tools/h5m/rest/TeamResource.java
```diff
@@ -0,0 +1,98 @@
...
+        teamService.delete(id);
+    }
+
+    @GET
+    @Path("{id}/members")
+    @Authenticated
+    @Operation(description = "List the members of a team")
+    public List<User> listMembers(@PathParam("id") long id) {
+        if (!identity.getRoles().contains(ADMIN_ROLE) && !userService.isMemberOf(id)) {
+            return List.of();
```
Arjun-Parmani: addmember and removeMember below returns 'ForbiddenException' . Why listMemebers is returning an empty list? Should also throw 'ForbiddenException'

src/main/webui/src/app/components/TeamMembersPanel.tsx
```diff
@@ -0,0 +1,167 @@
...
+              id="member-search"
+              size="lg"
+              labelText="Search Users"
+              placeholder="Search Users"
+              value={filter}
+              onChange={(e) => {
+                setFilter(e.target.value);
+              }}
+              onClear={() => {
+                setFilter('');
```
Arjun-Parmani: I think setFilter also should be (' ') right after the user adds the member to a team. Correct place should be onSucess(...) method when a user is successfully added.

src/main/webui/src/app/components/team/TeamMembersPanel.tsx
```diff
@@ -0,0 +1,167 @@
...
+        >
+          {nonMembers.map((user) => (
+            <ContainedListItem
+              key={user.id}
+              action={
+                <IconButton
+                  label="Add to Team"
+                  kind="ghost"
+                  onClick={() => {
+                    addMember.mutate({ path: { id: teamId, userId: user.id ?? 0 } });
```
Arjun-Parmani: Can a nonmember be a part of multiple teams? I can add 'alice' in performance and rhivos team both

barreiro: Yes, `alice` can be on multiple teams. Each team is a separate `addMember` call and `TeamMembersPanel` operates on whichever team is currently selected, so you'd repeat the action once per team. Authorization-wise, an admin can add her to any team, while a non-admin can only do it for teams they're themselves a member of.

Duplicate membership (`alice` being on the same team twice) is prevented by `TeamEntity.members` being a set, so adding the same user a second time is a no-op.

src/main/webui/src/app/components/TeamMembersPanel.tsx
barreiro: not quite. you can apply a filter and them add multiple users that match that filter without having to type it again.

src/main/java/io/hyperfoil/tools/h5m/entity/UserEntity.java
barreiro: Fixed by adjusting the `equals` / `hashcode` methods to something recommended for this type of JPA association.

PR #293, opened 2026-08-25
Notification UI

Arjun-Parmani:
Add notification config UI — create, edit, and delete notification configurations

Adds a Notification Config tab to the folder page with CRUD modals for managing notification configs (Webhook, Email, Slack, GitHub Issue).


barreiro:
I'm merging this one, but expect some changes. 

The notification config types should flow from backend to frontend, much like the detection nodes config. Then the config will be typed properly, and validated both on frontend and backend.

